Issue #6, opened 2018-09-17
What's up with fido_cred_set_fmt and fido_cred_set_type?


frozencemetery:
Hi, I've built a [toy client and server](https://github.com/frozencemetery/u2fsperiments) using libfido2.  It's working, but there are some behaviors I don't understand:

- If I don't set cred_fmt on my client, it defaults to fido-u2f.  (Okay, but contradicts man page.)
- If I don't set cred_fmt on the server, validation fails with INVALID_ARGUMENT (Annoying, but okay.)
- If I set cred_fmt on the server to fido-u2f, validation succeeds.  (Okay.)
- If I set cred_fmt on the client explicitly to fido-u2f, and set fido-u2f on the server, validation succeeds.  (Okay.)
- If I set cred_fmt to "packed", I still get a token with format fido-u2f, as reported by `fido_cred_fmt()`.  This passes server validation with server format set to fido-u2f.  (I don't understand this.)

Also:

- If I don't `fido_cred_set_type()` on the client, the client fails `make_cred` with invalid argument. (Annoying, but okay.)
- If I set both client and server to `COSE_ES256`, validation succeeds.  (Okay.)
- If I do `fido_cred_set_type()` only on the client to `COSE_ES256`, server validation succeeds.  (Unexpected, but okay.)
- If I `fido_cred_set_type()` on the client to `COSE_RS256`, the client fails `make_cred()` with invalid argument.  (I don't understand this.)
- If I `fido_cred_set_type()` on only the server to `COSE_RS256`, the validation succeeds, despite it not matching the `COSE_ES256` on the client.  (I really don't understand this.)

I'm running git master of libfido2, and have a U2F-only YubiKey from the GitHub promotion, if that matters.

Thanks!

frozencemetery:
Hi, thanks for the prompt reply!

> Perhaps it would make sense to have fido_dev_make_cred() return an
> error if the format chosen by authenticator does not match that of a
> previous fido_cred_set_fmt(). I would appreciate your thoughts on
> this.

I think it would make sense to have it fail rather than do something
unexpected.

In general, I'd prefer if the `fido_cred_set_fmt()` call failed - the
pattern of having `make_cred()` be the only meaningful failure means I
then have to go back with a debugger and figure out *which* argument was
INVALID.  But since this is a consistent pattern in the API, I assume
there's some obstacle to providing that kind of interface.

> What is happening right now is that, since attestation signatures are
[...]

PR #7, opened 2018-09-19
Clarify the role of type and fmt in man pages

frozencemetery:
[As promised, here's a PR for clarifying fmt and type - at least to me :) ]

Related: #6


martelletto:
Hi,

Thank you for the PR.

diff --git a/man/fido_cred_set.3 b/man/fido_cred_set.3
index add0319..b36b0dc 100644
--- a/man/fido_cred_set.3
+++ b/man/fido_cred_set.3
@@ -159,6 +159,9 @@ or
 A copy of
 .Fa fmt
 is made, and no references to the passed pointer are kept.
+Note that not all indicators support FIDO2 and therefore may not be
+able to generate
+.Vt "packed" .
 .Pp
 The
 .Fn fido_cred_set_type

This reads a bit strange to me. Are you sure you didn't mean 
authenticators?

diff --git a/man/fido_cred_verify.3 b/man/fido_cred_verify.3
index 2267f1f..ebee9b9 100644
--- a/man/fido_cred_verify.3
+++ b/man/fido_cred_verify.3
@@ -32,7 +32,8 @@ function verifies whether the client data hash, relying party ID,
 credential ID, and resident key and user verification attributes of
 .Fa cred
 have been attested by the holder of the private counterpart of
-the public key contained in the credential's x509 certificate.
+the public key contained in the credential's x509 certificate in the
+specified format using the indicated signature type.
 .Pp
 Please note that the x509 certificate itself is not verified.
 .Sh RETURN VALUES

fido_cred_set_type concerns the type of the credential to be 
created; it has no impact over the type of the key used by the 
authenticator to attest the created credential.

Perhaps the best way to communicate this would be to add the 
credential type to the list of attested attributes:

diff --git man/fido_cred_verify.3 man/fido_cred_verify.3
index 2267f1f..5c74156 100644
--- man/fido_cred_verify.3
+++ man/fido_cred_verify.3
@@ -29,7 +29,8 @@ A brief description follows:
 The
 .Fn fido_cred_verify
 function verifies whether the client data hash, relying party ID,
-credential ID, and resident key and user verification attributes of
+credential ID, type, and resident key and user verification
+attributes of
 .Fa cred
 have been attested by the holder of the private counterpart of
 the public key contained in the credential's x509 certificate.

Does that make sense to you?

-p.

frozencemetery:
It does.  Adjusted patch accordingly

martelletto:
merged; thank you!

-p.

Issue #10, opened 2018-12-07
libfido2 does not seem to work with BioPass FIDO2 authenticator


fgervais:
I'm trying to get pamu2fcfg to work with [this](https://www.ftsafe.com/onlinestore/product?id=37) fido2 authenticator.

Out of the box I'm getting the following:

```
fgervais@bionic_18041-yubico:~/pam-u2f$ pamu2fcfg -v -d -t es256
Setting origin to pam://bionic_18041-yubico
Setting appid to pam://bionic_18041-yubico
Setting user to fgervais
Setting user id to 8326c45fafb8d573c2444b33b2bd9284482ab739ba4238ba06844baa97dc00f1
get_report_descriptor: open
get_report_descriptor: open
get_report_descriptor: open
get_report_descriptor: open
get_report_descriptor: open
get_report_descriptor: open
tx: d=0x562b9c2361c0, cmd=0x86, buf=0x562b9c2361c0, count=8
  4f 51 da 5c 78 09 44 5d 
rx: initiation frame at 0x7ffdf6bbc370, len 64
  ff ff ff ff 86 00 11 4f 51 da 5c 78 09 44 5d 00 
  00 00 0b 02 01 00 01 0d 00 00 00 00 00 00 00 00 
  00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 
  00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 
tx: d=0x562b9c2361c0, cmd=0x90, buf=0x562b9c233dc0, count=180
  01 a4 01 58 20 b9 03 4e 0e db af 4a 3b f6 16 9a 
  1e fc 02 db c1 f2 c9 06 6a 6c 58 b7 08 52 ec 64 
  8b df a7 ff 37 02 a2 62 69 64 78 19 70 61 6d 3a 
  2f 2f 62 69 6f 6e 69 63 5f 31 38 30 34 31 2d 79 
  75 62 69 63 6f 64 6e 61 6d 65 78 19 70 61 6d 3a 
  2f 2f 62 69 6f 6e 69 63 5f 31 38 30 34 31 2d 79 
  75 62 69 63 6f 03 a2 62 69 64 58 20 83 26 c4 5f 
  af b8 d5 73 c2 44 4b 33 b2 bd 92 84 48 2a b7 39 
  ba 42 38 ba 06 84 4b aa 97 dc 00 f1 64 6e 61 6d 
[...]

PR #11, opened 2018-12-09
Encode COSE Algorithm as per spec

fgervais:
It seems like libfido2 is not correctly encoding small COSE algorithm values.

From what I understand, a device does not have the obligation to enforce those encoding rules and this is why the Yubikey 5 NFC does work even without this modification.

However the BioPass FIDO2 is stiffer and so does not allow for algorithm values < -25 encoded with an additional uint16_t.

This solves https://github.com/Yubico/libfido2/issues/10

Ref.:

FIDO 2.0: Client To Authenticator Protocol
  6. Message encoding

Integers must be encoded as small as possible.
  0 to 23 and -1 to -24 must be expressed in the same byte as the major type;
  24 to 255 and -25 to -256 must be expressed only with an additional uint8_t;

### DIFF
--- a/src/cbor.c
+++ b/src/cbor.c
@@ -327,6 +327,7 @@ encode_pubkey_param(int cose_alg)
 {
 	cbor_item_t		*item = NULL;
 	cbor_item_t		*body = NULL;
+	cbor_item_t		*alg_item = NULL;
 	struct cbor_pair	 alg;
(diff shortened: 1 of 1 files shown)


martelletto:
looks good to me; merging

thanks for the pull request!

-p.

PR #13, opened 2018-12-12
Add BioPass FIDO2 to udev rules

fgervais:
This is backported from libu2f-host.

Even though it's an easy thing, it's been done through

```
git show 15c6ca2 > udev.patch
patch -p1 -i udev.patch
```

to be on the safe side.


martelletto:
thank you; merging

-p.

Issue #20, opened 2019-03-21
cred.exe run failed


PhoenixBaymax:
I run windows/build.bat, and create vs2017 project and get cred.exe,but it crash.I found the reason is fclose(fp) in example/util.c,if I comment the statement,it run good,but it run good on Ubuntu.

martelletto:
Hi,

Which version of libfido2 are you running? There were a few changes 
in that area between the last released version (0.4.0) and HEAD.

-p.

PhoenixBaymax:
I tried again that use the latest version code.If I choose debug model, it work good,but if choose release model,the probelm occur that the libcbor and libress both are release lib.

martelletto:
Hi,

Thanks for testing. Just to confirm my understanding:

- The crash you were seeing is gone with HEAD;
- You are now experiencing a different problem when you set 
  CMAKE_BUILD_TYPE to 'Release'.

If these statements are correct, what is the problem you see when 
you set CMAKE_BUILD_TYPE to 'Release'?

-p.

PhoenixBaymax:
I git clone code in 2019/03/19,describe my problem with some pictures.
The picture of code:
![code](https://user-images.githubusercontent.com/5435863/54741266-ece64f80-4bf8-11e9-9752-ada8a31c2ba3.JPG)
debug .exe run in cmd:
![debug_model_run](https://user-images.githubusercontent.com/5435863/54741341-25862900-4bf9-11e9-90a7-66254a7b83c1.JPG)
release .exe run in cmd:
![release_model_run](https://user-images.githubusercontent.com/5435863/54741361-3171eb00-4bf9-11e9-98b7-2320725d2e4a.JPG)

By compairing the last pictures,you maybe unstander my problem!

martelletto:
Hi,

Thank you for the screenshots. I will try to reproduce the problem 
locally. Which version of Windows are you running?

-p.

PhoenixBaymax:
OS: Windows10 v1809 17763.379.
If u need my project built,I can send u.

martelletto:
Sure, that would be appreciated!

-p.

PhoenixBaymax:
what`s your email address?

martelletto:
pedro at yubico dot com

-p.

PhoenixBaymax:
I had sent email to u.

martelletto:
Hi,

I cannot reproduce the problem. As an extra data point, what happens 
[...]

PR #22, opened 2019-03-25
Add support for Windows shared lib

### LINKED ISSUE #21: No Shared Library Available for Windows
I noticed that support for Windows shared library is explicitly disabled in the CMake project.  I made some trivial changes locally and using this library as a shared library on Windows doesn't seem to cause any issues so why is it disabled?  If it is because of lack of familiarity with Windows builds then I would be happy to submit a PR for review.

The reason I want it as a shared library is because I wrote a set of .NET bindings to allow use of the library from .NET languages but the ability to link statically is not available to .NET projects.

borrrden:
Closes #21 

Also adds a proper build script for Windows


martelletto:
Thank you for the PR! One thing I noticed is that the statically linked examples, which used to work when linked against the debug runtime and to behave inconsistently when built with --config Release (issue #20), no longer work. When I try to execute them, I am presented with a dialog reading "The application was unable to start correctly (0xc000007b)", regardless of the runtime they are linked against. I plan to look further into this (and the PR as a whole) tomorrow. Perhaps you know why that's happening?

borrrden:
I am able to build and run (at least) manifest.exe fine using my script that I submitted.  I tweaked it a bit when it failed to behave properly on another machine and I realized that since you are already using export lists for other platforms I should simply be using them for MSVC as well.  That will be *a lot* cleaner.  I will update the PR but as for the other issue, it's more to do with the way the build is set up.  It would be better to have these two deps as CMake subdirectories so that they always get built with the correct configuration (more detailed comment on the issue itself)

borrrden:
P.S. I believe the particular problem you face could possibly be missing crypto-44.dll.  Windows is annoying about its search paths so even if it was found at link time it might not be found at runtime.  For cred.exe, in particular, the file needs to be adjacent to cred.exe

martelletto:
You are absolutely right; that was indeed the problem. Thank you!

PR #29, opened 2019-04-09
macos: relax manufacturer checking on fido devices

Silvercast:


Silvercast:
need some final tweaks...

PR #30, opened 2019-04-11
FIDO2 Standard 

jowin202:
According to CTAP2 Standard, Section 6 it is said

> The keys in every map must be sorted lowest value to highest. The sorting rules are:
> - If the major types are different, the one with the lower value in numerical order sorts earlier.
> - If two keys have different lengths, the shorter one sorts earlier;
> - If two keys have the same length, the one with the lower value in (byte-wise) lexical order sorts earlier

as it holds "id" < "type", the cbor map must contain id at first and then type. By changing this my FIDO2 Stick works.


martelletto:
merged; thanks!

PR #32, opened 2019-04-16
udev sync with libu2f-hosts; fix SecureClick typo

roycewilliams:
Mostly equivalent to https://github.com/Yubico/libu2f-host/pull/124

(I saw the note about libu2f-hosts being in maintenance mode and libfido2 being the active effort, so if I have other adds I'll PR them here in the future as well; should I do both? How long should the udevs be kept in sync?)

Also, it looks like that 'SeccureClick' typo has been the for quite some time - not sure why I'd missed it for so long!


martelletto:
Hi,

Thank you for the PR.

There is no need to file two PRs; while we transition from libu2f-host to libfido2, we will ensure the rules are synced. As to the PR itself, shouldn't we be using TAG+= instead of TAG=?

-p.

roycewilliams:
Yikes - artifact of my diff process, which is clearly broken. Fixed. That other PR was already accepted (and didn't have the + problem).

martelletto:
Merged. Thanks!

Issue #35, opened 2019-05-11
Incorrect device enumeration on Mac OS X


kobusgrobler:
function get_id() in hid_osx.c incorrectly filters out device because get_int32() returns 0xFFFFxxxx where xxxx is the actual VID/PID.
Ideally get_int16() should be used.

martelletto:
Fixed in f9da14560; thank you!

PR #36, opened 2019-05-11
Fix device enumeration issue for USB devices with VID/PID > 1FFF

kobusgrobler:
See issue #35


martelletto:
Merged; thank you! I took the liberty of amending your fix a little: I kept the check to prevent silent truncation from int32 when assigning vendor_id and product_id in fido_dev_info_t, but changed it from INT16_MAX to UINT16_MAX.

We should ideally switch fido_dev_info_vendor() and fido_dev_info_product() from int16_t to uint16_t as well, but that's an ABI-breaking change, so it will have to wait for libfido2 2.0.

PR #39, opened 2019-06-08
Avoid sign extension when printing USB vendor or device ID

emaste:
Cast vendor and device ID to uint16_t.  Previously vendor or device IDs
0x8000 or greater were displayed with a 0xffff prefix - for example,
product=0xffffa2ca.


martelletto:
Committed; thanks!

PR #55, opened 2019-08-17
add two new eWBM, and sort device IDs numerically

roycewilliams:
Keeping the eWBM device IDs sorted numerically, just like the other entries. 

I have both of the new devices (310 and 320) in the lab.


martelletto:
Merged; thanks!

PR #63, opened 2019-10-21
Add a fido_assert_sigcount() function

djm-google:
This allows access to the signature counter in an assertion statement
and is useful when implementing a standalone verifier.


martelletto:
Merging; thanks!

mfriedl:
i think the manage is still missing :)

martelletto:
@mfriedl implemented in b2567a9; thanks!

PR #82, opened 2019-11-27
Support compilers without packed structs

michaelforney:
I'd like to build libfido2 with a compiler that doesn't support packed structs, which are currently used by libfido2. Is this something you'd be willing to change?

I made some commits to use `uint8_t` arrays instead, and this doesn't seem to make the code more complex. I tested these changes using a U2F-only key and a CTAP2 key, and haven't noticed any issues so far.

I believe commit `cbor, u2f: avoid use of packed struct` also fixes a bug on big-endian systems where the credential ID length would be stored in the wrong byte of the attested credential data (it is currently marked with a `/* XXX */` comment).

Note: I understand this might take some time to review, or may not be accepted at all. I am just opening this PR as a proposal.


invd:
I'm an interested third party. 
It might be relevant for readers of this PR that there were [security issues](https://blog.inhq.net/posts/u2fhid_init_resp-information-leak/) in three products that used the [FIDO U2F 1.2 reference headers](https://fidoalliance.org/specs/fido-u2f-v1.2-ps-20170411/inc/u2f_hid.h) where the struct was not yet declared as `__packed__`. Summarized, the devices were leaking three bytes of discarded stack memory during each initial U2F connection handshake.

Replacing struct usage with "regular" buffers might be valid a valid step to reduce endianness-troubles and the related trade-offs during packet handling acceptable, but the resulting code should be tested thoroughly for e.g. similar information leaks in edge cases. Fortunately, the libfido2 libFuzzer setup with MSAN maintained by @martelletto should cover most of it already.

Out of interest: which compiler are you using?

martelletto:
I'd also be curious to learn more about your use case, if you don't mind sharing.

michaelforney:
My compiler is [cproc](https://git.sr.ht/~mcf/cproc), which deliberately stays very close to ISO C, and does not implement many GNU C extensions. It will implement C2X attribute syntax, but the only standard attributes are those that can be safely ignored:

> A strictly conforming program using a standard attribute remains strictly conforming in the absence of that attribute.

My goal is to build [oasis](https://github.com/oasislinux/oasis), my linux OS project which now includes libfido2 for security-key support in openssh. The current status of that effort is https://man.sr.ht/~mcf/cproc/doc/software.md#oasis.

Based on the changes in this PR, it doesn't seem like `__packed__` is offering much in terms of code simplicity or maintainability compared with working with the bytes directly.

martelletto:
Interesting projects; you have my admiration. As usual, C is complicated. Some of the currently packed structures don't strictly require the attribute. However, given their sensitive nature (as pointed out by @invd) and the fact C99 leaves it up to the compiler when/how to pad (the only guarantee it gives is that there can't be padding at the beginning of a structure), I would prefer if we could use packed structures with compilers that support it. That said, here is what I suggest:

- For each packed structure, provide a _pack() (serialize) and _unpack() (deserialize) function;
- Use these functions instead of hardcoded calls to memcpy()/fido_buf_{read,write}();
- Keep the existing behaviour if packed is supported;
- You could then keep a small patch in your tree, or merge the patch (implementing the body of the various _pack/unpack functions) with libfido2.

Does that make sense?

martelletto:
Closing after 30+ days of inactivity. Feel free to reopen, and I will be happy to revisit the PR.

PR #84, opened 2019-11-27
Add i/o to dev_info_t and register_dev_info_manfest function

James-ZHANG:
Add dev_io to dev_info so when a device is open custom i/o functions can be injected.
Add following API:
   - fido_dev_new_with_info(fido_dev_info *): initialize a device with dev_info returned by dev_info_manifest
   - fido_dev_open_with_info(fido_dev_info_t *, int ms): open a device and initialize the device i/o with that in fido_dev_info
   - fido_exit: properly frees any resource set by fido_init.
Internally,
   - add register_manifest functions so that custom device enumeration and i/o can be registered easily during fido_init, add hid_hidapi.c as an example usage where hidapi is used.


James-ZHANG:
Hi, pedro,
   I merged in the latest master here. Please tell me if you are OK with that and if not, I could switch to rebase approach.

martelletto:
Rebasing would be a lot better IMHO. Would you mind if I push some commits directly to your branch once it is rebased? I will also leave some comments above/below/somewhere in this UI. :)

CMakeLists.txt
```diff
@@ -88,6 +89,11 @@ else()
 		endif()
 	endif()
 
+	if (HIDAPI_FOUND)
+		set(HIDAPI_LIBRARIES hidapi-hidraw)
+		add_definitions(-DUSE_HIDAPI)
+	endif()
+
```
martelletto: Given the hidapi backend was merged in master, these bits should probably go.

examples/CMakeLists.txt
```diff
@@ -29,7 +29,11 @@ target_link_libraries(reset fido2)
 
 # cred
 add_executable(cred cred.c util.c ${COMPAT_SOURCES})
-target_link_libraries(cred fido2)
+if (HIDAPI_FOUND)
+	target_link_libraries(cred fido2 hidapi-hidraw)
+else()
+	target_link_libraries(cred fido2)
+endif()
```
martelletto: These too, I believe.

examples/cred.c
```diff
@@ -15,6 +15,9 @@
 #ifdef HAVE_UNISTD_H
 #include <unistd.h>
 #endif
+#ifdef USE_HIDAPI
+#include <hidapi/hidapi.h>
+#endif
```
martelletto: We don't appear to be using anything from this file in examples/cred.c.

examples/cred.c
```diff
@@ -299,5 +327,7 @@ main(int argc, char **argv)
 
 	fido_cred_free(&cred);
 
+	fido_exit();
```
martelletto: Probably no longer needed.

src/dev.c
```diff
@@ -195,26 +273,87 @@ fido_init(int flags)
...
+#else
+	fido_dev_register_manifest_func(fido_dev_info_manifest_linux);
+#endif // USE_HIDAPI
+#elif defined(OSX)
+	fido_dev_register_manifest_func(fido_dev_info_manifest_osx);
+#elif defined(WIN)
+	fido_dev_register_manifest_func(fido_dev_info_manifest_win);
+#elif defined(OPENBSD)
+	fido_dev_register_manifest_func(fido_dev_info_manifest_openbsd);
+#endif // LINUX|OSX|WIN|OPENBSD
```
martelletto: Wouldn't it better to have the various hid_*.c implementations define a hid_manifest() function each (given they are mutually exclusive) and just fido_dev_register_manifest_func(hid_manifest); here, avoiding the need for ifdefs?

src/dev.c
```diff
@@ -195,26 +273,87 @@ fido_init(int flags)
...
+	hid_exit();
+#endif
+	dev_manifest_func_node_t *curr = manifest_funcs;
+	dev_manifest_func_node_t *next = NULL;
+	while (curr != NULL) {
+		next = curr->next;
+		free(curr);
+		curr = next;
+	}
+	manifest_funcs = NULL;
```
martelletto: This should probably happen elsewhere. We may need to change the abstraction a bit, perhaps by introducing a function that does the reverse of fido_dev_register_manifest_func().

src/extern.h
```diff
@@ -134,4 +152,21 @@ int fido_verify_sig_rs256(const fido_blob_t *, const rs256_pk_t *,
...
+/* hid i/o */
+void *fido_hid_open(const char *);
+void fido_hid_close(void *);
+int fido_hid_read(void *, unsigned char *, size_t, int);
+int fido_hid_write(void *, const unsigned char *, size_t);
+
+/* device manifest registration */
+typedef int (*dev_manifest_func_t)(fido_dev_info_t *dev_infos, size_t ilen,
+                                   size_t *olen);
+int fido_dev_register_manifest_func(const dev_manifest_func_t func);
```
martelletto: Some of the additions appear to be duplicates.

James-ZHANG: No, I don't think these are defined elsewhere.

James-ZHANG:
> Rebasing would be a lot better IMHO. Would you mind if I push some commits directly to your branch once it is rebased? I will also leave some comments above/below/somewhere in this UI. :)

I rebased against the latest mater and did a force push, and also resolve all your comments. Please check the diffs of the latest commits since all commits in this history page are now outdates.
I'm going to add you as a collaborator so feel free to push directly.

martelletto:
Merged; thanks!

PR #88, opened 2019-11-29
fix attcred_raw.id_len on big-endian systems

michaelforney:
This is a standalone fix for the endianness issue mentioned in #82.

The attested credential data uses big-endian encoding for the `credentialIdLength`:
https://www.w3.org/TR/webauthn/#sec-attested-credential-data


martelletto:
Thanks! Merging.

PR #89, opened 2019-12-02
make headers self-contained on linux

James-ZHANG:


src/fido.h
```diff
@@ -14,23 +14,7 @@
...
-	fido_dev_io_read_t  *read;
-	fido_dev_io_write_t *write;
-} fido_dev_io_t;
-
-typedef enum {
-	FIDO_OPT_OMIT = 0, /* use authenticator's default */
-	FIDO_OPT_FALSE,    /* explicitly set option to false */
-	FIDO_OPT_TRUE,     /* explicitly set option to true */
-} fido_opt_t;
+#include "common.h"
```
martelletto: This prevents inclusion of fido.h once the header files are installed. If you do:

$ (cd build && cmake -DCMAKE_INSTALL_PREFIX=/tmp/root ..) && make -C build -j8 install
$ echo '#include <fido.h>' | cc -I/tmp/root/include -xc -c -

There are similar problems in fido/bio.h, fido/credman.h, fido/eddsa.h, and fido/es256.h.

martelletto:
I am leaning towards moving all type definitions (including types.h) to fido/types.h instead of rolling fido/common.h. What do you think? I have pushed a tentative branch (along with a script to check self-containment) which contains this PR: https://github.com/Yubico/libfido2/tree/self_contained_headers. I haven't had time to review it yet; my apologies.

include_check
```diff
@@ -0,0 +1,16 @@
...
+			cc $CFLAGS -Isrc -xc -c - -o /dev/null 2>&1
+		echo $f $CFLAGS $?
+	done
+}
+
+check examples
+check fuzz
+check openbsd-compat
+CFLAGS=-D_FIDO_INTERNAL check src
+check src/fido.h
```
James-ZHANG: I think src/fido/*.h should be checked as well, no?

James-ZHANG:
Thanks, pedro. I pulled your branch here and made some comments. It looks good to me and I think making fido/types.h public is a good idea.

include_check
martelletto: You are right. Fortunately, src/fido/*.h is fine.

martelletto:
Thanks. Merging.

PR #91, opened 2019-12-03
integrate hidapi by adding src/hid_hidapi.c

James-ZHANG:
Hi, pedro,
   This PR is a preliminary to #84 . It adds `src/hid_hidapi.c`, which uses hidapi to provide the low-level fido_dev_io_* function implementations. During build time, if `-DUES_HIDAPI=1` is passed to `cmake` then `hidapi-raw` would be searched and if found, `hid_hidapi.c` would be linked into the library.

   This is my first time trying to add licence texts into a source file. Please correct me if I do it wrong:)


src/dev.c
```diff
@@ -195,6 +198,18 @@ fido_init(int flags)
 {
 	if (flags & FIDO_DEBUG || getenv("FIDO_DEBUG") != NULL)
 		fido_log_init();
+#ifdef USE_HIDAPI
+	if (hid_init() != 0)
+		errx(1, "hid_init failed");
```
martelletto: Having the library call errx() and exit the application is not ideal. If you are worried about hid_init() failing, then I suggest keeping a static variable in hid_hidapi.c to track whether hid_init() has been called, and returning an error from there.

src/hid_hidapi.c
```diff
@@ -0,0 +1,121 @@
+/* hidapi license text
+Copyright (c) 2010, Alan Ott, Signal 11 Software
```
martelletto: The copyright is normally the author's; if you wrote the code for work, the copyright is usually assigned to your employer (as @djm-google did in hid_openbsd.c).

src/dev.c
James-ZHANG: According to hidapi.h's documentation hid_init and hid_exit calls are not necessary unless multi-threading is involved: https://github.com/signal11/hidapi/blob/master/hidapi/hidapi.h#L82.
Thus I removed these calls altogether, a quick test of examples/cred on my Linux workstation at least works.

martelletto:
Thanks! Merging.

PR #94, opened 2019-12-09
adding a static code analysis github workflow

nevun:
This only runs on push for now so PRs are not covered. PRs are special in that if it comes from a fork it does not have access to the docker registry secrets in the repo but I have found no way of testing for the presence from within a github flow.

As of today the scanners show no warnings.


martelletto:
thanks; merging

PR #96, opened 2019-12-12
Allow to add customized logging functions via fido_init().

ikimonogakari:
I am adding this because when we plug this library to some main process, we want to be able to reuse the logging function that main process uses. So that we have the log at the same place.

### DIFF
--- a/src/dev.c
+++ b/src/dev.c
@@ -243,6 +243,9 @@ fido_dev_info_manifest(fido_dev_info_t *devlist, size_t ilen, size_t *olen)
 
 	*olen = 0;
 
+        if (fido_dev_register_manifest_func(fido_hid_manifest) != FIDO_OK)
+		return (FIDO_ERR_INTERNAL);
+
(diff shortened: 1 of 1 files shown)


ikimonogakari:
Hi, Pedro,
Would you mind taking a look at this change?
This should be a simple change, which enables users to add their own logging functions through fido_init().

Thank you,
Cary

martelletto:
Hi Cary,

Thank you for the PR! Sure, I will give it a look early next week.

-p.

martelletto:
Hi Cary,

Changing the number of parameters taken by fido_init() requires a major API bump. We could make it a minor bump by adding a new fido_set_logger() symbol instead.

-p.

ikimonogakari:
Thank you Pedro for your advice.
I have changed this PR with your suggestion.

martelletto:
Thanks. Could you rebase on master and squash your commits?

PR #103, opened 2019-12-20
Record keyid for checkonly

James-ZHANG:
Hi, Pedro,
   This change is intended to let fido_dev_get_assert records the key_id in the attstmt_t for check-only cases as well. We believe this is a common use case (a user has a sequence of key_ids from the rp and they want to check which one of them is valid for the device, without a user touch).


martelletto:
Hi James,

Nice catch; thank you! Let me know if you disagree with my tweaks above. I will otherwise test the changes (probably on Monday) and merge the PR.

-p.

PR #105, opened 2019-12-27
log init frame's data length instead of a useless struct size

James-ZHANG:


src/io.c
```diff
@@ -179,9 +179,10 @@ rx(fido_dev_t *d, uint8_t cmd, unsigned char *buf, size_t count, int ms)
 		fido_log_debug("%s: rx_preamble", __func__);
 		return (-1);
 	}
+	flen = (f.body.init.bcnth << 8) | f.body.init.bcntl;	
 
 	fido_log_debug("%s: initiation frame at %p, len %zu", __func__,
-	    (void *)&f, sizeof(f));
+	    (void *)&f, sizeof(flen));
```
martelletto: Hi James,

Did you mean to log the value of `flen` instead of its size?

-p.

James-ZHANG: I do. Mistakes fixed.

martelletto:
Thanks! Merging.

PR #111, opened 2020-01-09
let fido_tx handle properly for empty payloads as well

James-ZHANG:
This is not a requirement for libfido2 since standard fido2/u2f commands never have a non-empty payloads. Yet I think it is good to add the proper handling to prevent future suprise.

### DIFF
--- a/src/io.c
+++ b/src/io.c
@@ -33,15 +33,37 @@ struct frame {
 #define MIN(x, y) ((x) > (y) ? (y) : (x))
 #endif
 
-static ssize_t
+static int
+tx_empty(fido_dev_t *d, uint8_t cmd)
+{
+	struct frame	*fp;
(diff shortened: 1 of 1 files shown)


martelletto:
Hi James,

Nice catch. The existing code is particularly confusing given tx_preamble() checks `count != 0` before calling memcpy() and proceeds to return zero (if `count == 0`) to indicate success. There is also the nonsensical check for `cmd & 0x80 != 0` followed by `0x80 | cmd`, but we can leave that for another day.

I'd like to avoid `ssize_t` if possible, though. IMHO, `ssize_t` is a bit of a hack and is likely to cause more problems than it solves. I have pushed an alternative approach. Please let me know what you think.

-p.

James-ZHANG:
Thanks, Pedro, I agree with the alternative approach.

PR #112, opened 2020-01-15
src/log.c: bump XXDLEN 16->32 to ensure buffer isn't too small

dtzWill:
GCC 8 and GCC 9 both are uncomfortable with the snprintf'ing
of format string "%04zu: %02x" fitting into 16 bytes,
and it's unclear to me this is an unwarranted concern.

If the buffer is indeed too small, the result would not
be a memory safety error AFAIK but the truncated string
would provide poor log messages.

Likely harmless but instead of worrying about any of that
just allocate a few more bytes to make it clear to programmers and
compilers alike that there's nothing to worry about.

### DIFF
--- a/src/log.c
+++ b/src/log.c
@@ -13,7 +13,7 @@
 
 #ifndef FIDO_NO_DIAGNOSTIC
 
-#define XXDLEN	16
+#define XXDLEN	32
 #define XXDROW	128
 #define LINELEN	256
 


martelletto:
Hi,

That’s odd. I cannot reproduce the issue with gcc-9 (Ubuntu 9.1.0-2ubuntu2~19.04) 9.1.0 or gcc-8 (Ubuntu 8.3.0-6ubuntu1) 8.3.0.  I am not opposed to cranking XXDLEN , but I’d like to understand what’s going on. What is the warning you see, and which version/flavour of GCC 8 and GCC 9 are you using?

Thanks!

-p.

dtzWill:
On Thu, 16 Jan 2020 06:24:59 -0800, pedro martelletto <notifications@github.com> wrote:
> Hi,
> 
> That’s odd. I cannot reproduce the issue with gcc-9 (Ubuntu 9.1.0-2ubuntu2~19.04) 9.1.0 or gcc-8 (Ubuntu 8.3.0-6ubuntu1) 8.3.0.  I am not opposed to cranking XXDLEN , but I’d like to understand what’s going on. What is the warning you see, and which version/flavour of GCC 8 and GCC 9 are you using?

Oh, I'm on NixOS for this-- I was using gcc 8.3.0 and gcc 9.2.0.

It's likely the issue is that we (like most distributions) enable
various compiler flags by default for some security hardening.

In particular the flags we enable are these:

`-Wformat`
`-Wformat-security`
`-Werror=format-security`

Sorry, I forget these are on sometimes O:).

Hope this helps with reproducing!

LMK if that doesn't work, or whatever you're able to sort out once you
can reproduce the warning.

Thanks!

martelletto:
Hi,

My apologies; I was jet-lagged beyond recognition. In fact I still am, but here we go. :)

I know what is going on: given the format string, gcc thinks (quite correctly) that the buffer needs to be at least 25 bytes long; 20 for %04zu, 2 for ': ', 2 for %02x on a char, plus nul termination. What I can't explain is why you see the warning and I don't, even with -Wformat -Wformat-security -Werror=format-security. I suggest we crank XXDLEN to 32 (as implemented in the PR), XXDROW, and possibly LINELEN as well.

I will read the gcc docs once I've had some sleep to see if I can spot the -W flag responsible for this warning.

Thank you!

-p.

martelletto:
It seems to be -Wformat-truncation=2.

martelletto:
Merging. Thanks!

PR #117, opened 2020-02-08
headers: Add 'extern "C"' declarations for C++ support.

conte91:
Linking libfido2 into a C++ project will fail, as the headers
are missing the 'extern "C"' declarations - thus, the linker will
assume C++ name wrangling and fail to find the invoked functions into
libfido2.so. Add extern "C" to all declarations inside header files.


martelletto:
merging; thanks!

PR #126, opened 2020-02-21
Use GNUInstallDirs to install documentation and manpages

tgurr:
Use GNUInstallDirs to install the documentation and manpages which allows
to install them into their arch independent locations instead of always
ending up under prefix, which can be e.g. /usr/x86_64-pc-linux-gnu on
cross environments.

Also install the html documentation in the html subdirectory under docdir.


martelletto:
merging; thanks!

PR #128, opened 2020-02-22
Add example for OSX 

dirkx:
Add example for OSX where the path is a bit non-intuitive (and has spaces in it).


martelletto:
Slightly different fix committed in 725ba0c and 42d876f6. Thank you!

PR #130, opened 2020-02-23
Only use -fstack-protector-all if it is supported

nbraud:
This flag isn't available on several architectures, including alpha, hppa, and
ia64, resulting in [build failures] on several Debian ports.

[build failures]: https://buildd.debian.org/status/package.php?pkg=libfido2


martelletto:
merging; thanks!

PR #133, opened 2020-02-25
Update FreeBSD manpage location

sunpoet:
LOCALBASE/share/man is a valid path for manpages since [r523104](https://svnweb.freebsd.org/changeset/ports/523104).


martelletto:
Merging. Thanks!

PR #134, opened 2020-02-25
fix broken install paths of html man pages

James-ZHANG:
Hi, Pedro,
    This is the first time I touched debian packaging stuff so please correct me if I am wrong.
    I tried building the debian source packages under debian/ on my workstation, but it failed at the `dh_install` step. After I made the change in this commit the install worked smoothly.
    I guess at some point you changed the installation path of these html files and forgot to upgrade those?


martelletto:
Hi James,

libfido2 was recently packaged by Debian: https://salsa.debian.org/auth-team/libfido2/-/tree/debian/sid/debian. I recommend using their packaging bits as a reference point. In the long term, we will probably drop debian/ from libfido2.

As far as the PR is concerned, you are right, this changed in 8aec3e0. I will merge the fix. Thanks!

-p.

Issue #132, opened 2020-02-25
Request to bring libfido over to MSYS


cw789:
Would it be possible for anyone to bring libfido over to [MSYS2](https://github.com/msys2/MSYS2-packages).

I think a lot of people are using OpenSSH on Windows via "Git for Windows".
With having libfido2 within MSYS it would be currently the easiest way for using the new OpenSSH FIDO/U2F support.

Related Issue: https://github.com/git-for-windows/git/issues/2525

riedel:
I had no problem actually compiling it with mingw on msys.  Also there is a cygwin package. So it would be up to the msys folks to simply package it. Or did you have any issues?

tavrez:
@riedel could you please explain your build steps? I can't build it using msys

riedel:
`cmake -G "Unix Makefiles" ..` works for me in the mingw64 shell.

The library (not the tools) itself also build under MSYS natively if you disable the -Wall and pedantic check (no clue what is the problem with string.h here) in the cmake Makefile
with `CMAKE_LEGACY_CYGWIN_WIN32=1 cmake ..` 

But don't you simply post your errors here?

riedel:
As said there is a cygwin packages with a patch to CMakeLists.txt that theoretically could be merged into the source here

tavrez:
@riedel Thank you, the problem was with -std=c99, it is now working.

martelletto:
Fixed in 5a9be2f0 by @riedel.

PR #137, opened 2020-02-27
Add MSYS and CYGWIN build support

riedel:
#132 asked for it. So this is my take.

Actually I found some interesting quirks, that could be problematic in other cases as well, e.g. using _NSIG instead of NSIG.

And that defining -std=c99 you don't get some some functions like strdup, so I defined _DEFAULT_SOURCE as I guess this does not hurt to make sure they are in...


martelletto:
Thanks! Merging.

cw789:
Thanks a lot Till.

Issue #129, opened 2020-02-23
Check for vendor extensions and generate clear error messages


riedel:
From the readme I get that libfido is vendor indepent.

The spec states: `Command codes in the range between authenticatorVendorFirst and authenticatorVendorLast may be used for vendor-specific implementations. For example, the vendor may choose to put in some testing commands. Note that the FIDO client will never generate these commands. All other command codes are reserved for future use and may not be used.`

Currently the following lines use yubikey specific codes without any precautions (e.g. checking vendor):
https://github.com/Yubico/libfido2/blob/393c18ccc24b71b30340deeebb92f32e4c6b4a42/src/credman.c#L140
https://github.com/Yubico/libfido2/blob/38c0a50f70edb41ecd25be0e4a6fd63fce9fea9b/src/bio.c#L112

Documentation/Specs seems to be missing

The yubikey vendor extentions seem to be suffixed _PRE.

I would suggest a configure switch for dis/enabling those and dependent code.

martelletto:
Hi,

It is up to the application, based on metadata obtained through [fido_dev_get_cbor_info()](https://developers.yubico.com/libfido2/Manuals/fido_dev_get_cbor_info.html), to understand the commands and functionalities supported by an authenticator, and to decide which commands to issue, how, and when. If we move this logic to the library, libfido2 will no longer be vendor independent.

Regarding credential management and the biometric enrollment API, these are features of FIDO 2.1 preview keys. Unfortunately, the unreleased FIDO specs are not publicly available.

-p.

riedel:
So does that mean that  FIDO 2.1  will use the vendor_keys 0x40 and 0x41. That seems rather strange to me. 

Why would a vendor independent library hard code vendor keys of one vendor? My comment was exactly that the mentioned lines or rather https://github.com/Yubico/libfido2/blob/780ad3c258aea5028b7b94c6623a96da3fd55224/src/fido/param.h#L34 breaks vendor independence.

martelletto:
Hi,

> So does that mean that FIDO 2.1 will use the vendor_keys 0x40 and 0x41. That seems rather strange to me.

That means that FIDO 2.1 preview (FIDO_2_1_PRE) uses values 0x40 and 0x41.

> Why would a vendor independent library hard code vendor keys of one vendor? 

You seem to extrapolate from "undocumented" to "one vendor". I completely understand your frustration, but I am afraid that is not the case. FIDO_2_1_PRE is implemented by devices from multiple vendors.

[...]

PR #138, opened 2020-02-27
Warn fido21 pre

riedel:
Since I complained about the documentation in #129 , maybe it makes sense to add a least a word of warning that the functions are not included in the current spec.

### DIFF
--- a/man/fido_bio_dev_get_info.3
+++ b/man/fido_bio_dev_get_info.3
@@ -34,8 +34,9 @@
 .Sh DESCRIPTION
 The functions described in this page allow biometric
 templates on a FIDO2 authenticator to be listed, created,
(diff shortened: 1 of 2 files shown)


martelletto:
Hi,

I've made some changes. Do they make sense?

Thank you,

-p.

riedel:
I like it. Thanks, would have same me a lot of googling (hopefully someone else to) and maybe makes people aware that it could be a mess when structures or command codes switch.

Not so important side note: Since you nicely introduced CAVEATS, I was only wondering if the "Please note..." part should not also go there. (where it will likely be overlooked :) )

martelletto:
Thank you for the PR and feedback!

Issue #147, opened 2020-03-11
Incorrect message from fido2-token


rgerganov:
I believe `fido2-token -I -c <dev>` should print the number of existing and the number of *remaining* RKs. Right now it says this:
```
$ fido2-token -I -c /dev/hidraw6
Enter PIN for /dev/hidraw6: 
existing rk(s): 2
possible rk(s): 48
```
OpenSSH got this right:
```
$ ssh-add -Kv 
Enter PIN for security key: 
...
debug1: read_rks: existing 2, remaining 48
...
```

PR #148, opened 2020-03-11
tools: fix the message for resident credentials metadata

rgerganov:
The message should say "remaining" instead of "possible" RKs.

Fixes issue #147


martelletto:
Thank you!

PR #151, opened 2020-04-07
remove several unnecessary fuzz wrappers

James-ZHANG:
The EC_KEY_get0_* and the two handle functions have deterministic behaviors---they return successful values for valid inputs, and error values otherwise. I don't think these functions should be wrapped to mock probabilistic failures. Indeed, our fuzzing tool has generated false-negative reports due to these mocking.

### DIFF
--- a/fuzz/wrap.c
+++ b/fuzz/wrap.c
@@ -232,22 +232,6 @@ WRAP(EC_KEY *,
 	1
 )
 
-WRAP(const EC_GROUP *,
-	EC_KEY_get0_group,
-	(const EC_KEY *key),
-	NULL,
-	(key),
-	1
-)
-
-WRAP(const BIGNUM *,
-	EC_KEY_get0_private_key,
-	(const EC_KEY *key),
-	NULL,
-	(key),
-	1
-)
-
 WRAP(EC_POINT *,
 	EC_POINT_new,
 	(const EC_GROUP *group),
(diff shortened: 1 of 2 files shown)


martelletto:
Hi James,

You have startled my curiosity. :) If you don't mind sharing, how did the wrapping of these symbols lead to false-negative reports by your fuzzing tool?

-p.

James-ZHANG:
Hi, Pedro,

After reviewing the failure I realized the problem was not due to `libfido2`, but because the fuzz tooling in Google wrapped these symbols for all dependency libraries, including OpenSSL. The false negatives were triggered from these dependencies since not all of them guard against failures of these functions.

That said, IMHO a function need not be mocked if its behavior is fully decidable by the "program runtime state". A function needs such mocking only if it depends on non-program-controlled stuff like I/O, thread scheduling, memory allocation, etc..

This decision is of course a matter of personal flavor, so if you disagree, just feel free to close this PR, and I'll reconfigure my own fuzz tooling.

martelletto:
Hi James,

I tend to disagree, but would be very much interested in hearing your thoughts, if you don't mind burning a few cycles with me. My rationale when writing the code was: these functions are deterministic _now_, but there is no guarantee they will remain so. Since we're talking about a possible NULL pointer dereference in C, we should err on the side of caution and check their return values. If we check the return values, we ought to have a way of fuzzing the fail path, hence the wrapping.

On a side note, I feel strongly about being a burden on others, so please accept my sincere apologies for the time spent on debugging the side-effects incurred by the wrapping of these symbols.

-p.

martelletto:
Closing after touching base with James.

PR #152, opened 2020-04-07
add surrouding BN_CTX_start|end calls for BN_CTX_get

James-ZHANG:
I believe these surrounding calls should be made per [OpenSSL's manpages](https://www.openssl.org/docs/manmaster/man3/BN_CTX_get.html).


martelletto:
You are right. Thanks!

PR #160, opened 2020-04-18
call gzip with -n argument for manpages

prusnak:
calling gzip without -n breaks reproducible builds
see the following for more info:

https://wiki.debian.org/ReproducibleBuilds/TimestampsInGzipHeaders


martelletto:
Thanks!

eli-schwartz:
Note also that compressing manpages can simply be skipped. Linux package managers generally have builtin options to compress manpages, which already call the compressor with the right (reproducible) options, and are sometimes tuneable (users can decide whether they prefer uncompressed manpages, or choose a diffferent compression method like bzip2).

PR #167, opened 2020-05-03
MSVC Fixes

jacquesg:
This addresses 2 problems that I've found:

* ```__attribute__``` isn't available, use ```__declspec``` instead
* The ```uid_t``` and ```mode_t``` types conflict for example with Perl's typedefs. Other projects also seem to chose the types I suggest:

References:
* https://gitlab.kitware.com/cmake/cmake/-/blob/001043ac3078c49651f6af0f1ff1b31ef71a7665/Source/cmStandardIncludes.h#L56
* https://fossies.org/linux/perl/win32/win32.h
* https://github.com/llvm-git-prototype/llvm/blob/master/lldb/include/lldb/Host/windows/PosixApi.h#L76

### DIFF
--- a/openbsd-compat/types.h
+++ b/openbsd-compat/types.h
@@ -23,13 +23,6 @@ typedef uint32_t        uid_t;
 #endif
 
 #ifdef _MSC_VER
-typedef unsigned char   u_char;
-typedef unsigned short  u_short;
-typedef unsigned int    u_int;
-typedef uint32_t        in_addr_t;
-typedef uint32_t        mode_t;
-typedef uint32_t        uid_t;
-
 #include <basetsd.h>
 typedef SSIZE_T ssize_t;
 
--- a/src/fido.h
+++ b/src/fido.h
@@ -32,6 +32,12 @@
 extern "C" {
 #endif /* __cplusplus */
 
+#ifdef _MSC_VER
+#define FIDO_DEPRECATED(reason) __declspec(deprecated(reason))
+#else
+#define FIDO_DEPRECATED(reason) __attribute__((__deprecated__(reason)))
+#endif
+
 fido_assert_t *fido_assert_new(void);
 fido_cred_t *fido_cred_new(void);
 fido_dev_t *fido_dev_new(void);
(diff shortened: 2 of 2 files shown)


openbsd-compat/types.h
```diff
@@ -27,8 +27,8 @@ typedef unsigned char   u_char;
 typedef unsigned short  u_short;
 typedef unsigned int    u_int;
 typedef uint32_t        in_addr_t;
-typedef uint32_t        mode_t;
-typedef uint32_t        uid_t;
+typedef unsigned short  mode_t;
+typedef long            uid_t;
```
martelletto, line 31: Given we don't use these definitions from openbsd-compat/types.h, would you agree it is better to just trim them?

jacquesg: ```mode_t``` is actually used by the ```fuzz/``` directory. The others are not. Ideally these shouldn't make its way directly or indirectly into the public headers.

martelletto: `mode_t` is used in `fuzz/`, but it isn't coming from openbsd-compat/types.h, as far as I can see.

jacquesg: Its is included indirectly:

```
fuzz_assert.c
24:#include "../openbsd-compat/openbsd-compat.h"

fuzz_bio.c
20:#include "../openbsd-compat/openbsd-compat.h"

fuzz_cred.c
19:#include "../openbsd-compat/openbsd-compat.h"

fuzz_credman.c
20:#include "../openbsd-compat/openbsd-compat.h"

fuzz_mgmt.c
18:#include "../openbsd-compat/openbsd-compat.h"
```

```openbsd-compat.h``` includes ```types.h```:
```
#if defined(_MSC_VER)
#include "types.h"
#endif
```

martelletto: Indeed, but the `fuzz/` files including openbsd-compat.h (and therefore types.h) don't use `mode_t`. The only files in `fuzz/` using `mode_t` (preload-snoop.c and preload-fuzz.c) don't include anything but system headers, which means there is nothing using the definition from types.h, as far as I can see. Does that make sense?

jacquesg: Indeed. I will remove these then, its a better fix.

jacquesg: All fixed.

martelletto:
Thank you! Merging.

Issue #165, opened 2020-05-02
Implement support for HID report sizes != 64 bytes


FabianHenneke:
**Describe the feature you would like.**
The U2F/CTAPHID specs allow FIDO devices to use any HID report size, not just the very common value 64 bytes. The report size can be extracted from the HID report descriptor, which has to be parsed anyway to determine whether the device declares the FIDO usage (page).

Currently, libfido2 assumes that all FIDO devices understand HID reports of size 64 bytes, which e.g. is not the case with [WearAuthn](https://github.com/FabianHenneke/WearAuthn) and in general limits the (underexplored) design space for Bluetooth HID authenticators. I would like to see libfido2 become fully spec compliant and support any (sensible) report size, as do the Windows 10 and Chrome implementations.

**Describe the impact to existing functionality.**
This new feature can be implemented in a backwards compatible manner. If the HID report descriptor cannot be parsed correctly, the implementation would fall back to the default 64 bytes. Fuzzing and tests could also easily be extended to cover arbitrary report sizes.

**Describe how you are willing to help.**
If it is reasonably likely that a PR adding this functionality, accompanied by tests and fuzzer integration, will be accepted, I would gladly start working on it.

martelletto:
Hi,

CC'ing @James-ZHANG, who might be interested in this.

A PR implementing this feature, in the lines described (tests, fuzzer integration), and without adding much complexity to existing code, would definitely be welcome. I unfortunately lack the time to work on it myself, but please count on me for help and support.

-p.

FabianHenneke:
I found code style and in particular use of whitespace to be rather inconsistent between (and sometimes within) files. Is there an official policy on this or should I try to preserve the existing local structure as far as possible?

martelletto:
Hi,

We try to abide by https://man.openbsd.org/style.

-p.

martelletto:
Implemented in #172, #173, #180, and #182.

PR #168, opened 2020-05-05
Extract and use report lengths contained in HID report descriptor

FabianHenneke:
This PR is meant to address #165 by replacing the hardcoded HID report length of 64 bytes by the lengths reported in the FIDO devices HID report descriptor.

It is broken up into an initial commit that replaces the hardcoded HID report length with configurable input and output report lengths, while still using the default report length in all backends. This is meant to simplify the review process. This commit also adapts the regression tests and fuzzers accordingly.

The later commits implement the extraction of the report lengths from the report descriptor for the various backends. Here, the status of the individual commits is as follows:

- [x]  Windows: Builds and performs as expected as confirmed by a manual test with both a 64-byte and a 62-byte report FIDO device.
- [x] Linux: Builds and performs as expected as confirmed by a manual test with both a 64-byte and a 62-byte report FIDO device.
- [ ] macOS: Does build in the CI, but I haven't been able to test it since I do not have access to a physical device running macOS.
- [ ] hidapi: Is fundamentally incompatible with devices with unknown HID report sizes since it does not extract the report lengths from the report descriptor on all platforms. The respective commit adds a note about this, but this may warrant mentioning in the README (or elsewhere).
- [ ] OpenBSD: Not implemented yet. I lack experience with OpenBSD and do not know which library I should use to fetch the HID report descriptor or report lengths for both USB and Bluetooth HID devices.

### DIFF
--- a/src/fido/types.h
+++ b/src/fido/types.h
@@ -206,13 +206,20 @@ struct fido_ctap_info {
 	uint8_t  flags;    /* capabilities flags; see FIDO_CAP_* */
 })
 
+typedef struct fido_dev_io_info {
+	void	*io_handle;		/* opaque i/o handle */
+	uint16_t report_in_len;		/* length of the HID input report */
+	uint16_t report_out_len;	/* length of the HID output report */
+} fido_dev_io_info_t;
+
 typedef struct fido_dev {
 	uint64_t              nonce;     /* issued nonce */
 	fido_ctap_info_t      attr;      /* device attributes */
 	uint32_t              cid;       /* assigned channel id */
 	char                 *path;      /* device path */
-	void                 *io_handle; /* abstract i/o handle */
+	fido_dev_io_info_t   *io_info;   /* extra info needed to perform i/o */
 	fido_dev_io_t         io;        /* i/o functions */
+	bool                  legacy_io; /* i/o relies on default report size */
 	fido_dev_transport_t  transport; /* transport functions */
 } fido_dev_t;
 
--- a/src/dev.c
+++ b/src/dev.c
@@ -106,13 +106,50 @@ find_manifest_func_node(dev_manifest_func_t f, dev_manifest_func_node_t **curr,
 	}
 }
 
+static fido_dev_io_info_t *
+fido_dev_io_open_wrapper(fido_dev_t *dev, const char *path) {
+	fido_dev_io_info_t *io_info;
+
+	if (dev->legacy_io) {
+		if ((io_info = malloc(sizeof(*io_info))) == NULL)
+			return (NULL);
+
+		if ((io_info->io_handle = dev->io.open(path)) == NULL) {
+			fido_log_debug("%s: dev->io.open", __func__);
+			free(io_info);
+			return (NULL);
+		}
+
+		/*
+		 * Use standard report lengths for backwards compatibility with
+		 * custom IO functions enabled with fido_dev_set_io_functions.
+		 */
+		io_info->report_in_len = MAX_CTAP_REPORT_LEN;
+		io_info->report_out_len = MAX_CTAP_REPORT_LEN;
+
+		return io_info;
+	} else {
+		return dev->io.open(path);
+	}
+}
+
+static void
+fido_dev_io_close_wrapper(fido_dev_t *dev) {
+	if (dev->legacy_io) {
+		dev->io.close(dev->io_info->io_handle);
+		free(dev->io_info);
+	} else {
+		dev->io.close(dev->io_info);
+	}
+}
+
 static int
 fido_dev_open_tx(fido_dev_t *dev, const char *path)
 {
 	const uint8_t cmd = CTAP_CMD_INIT;
 
-	if (dev->io_handle != NULL) {
-		fido_log_debug("%s: handle=%p", __func__, dev->io_handle);
+	if (dev->io_info != NULL) {
+		fido_log_debug("%s: handle=%p", __func__, (void *)dev->io_info);
 		return (FIDO_ERR_INVALID_ARGUMENT);
 	}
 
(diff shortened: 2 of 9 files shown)


martelletto:
Thank you! Much appreciated. I am planning to do a formal review over the weekend, but will try my best to provide feedback of some kind before then. Out of curiosity, are the changes enough for WearAuthn to work on Linux (as I understand it, hidraw(4) abstracts both USB and Bluetooth HID) for SSH auth?

FabianHenneke:
> Thank you! Much appreciated. I am planning to do a formal review over the weekend, but will try my best to provide feedback of some kind before then. 

Thank you very much!

> Out of curiosity, are the changes enough for WearAuthn to work on Linux (as I understand it, hidraw(4) abstracts both USB and Bluetooth HID) for SSH auth?

Thanks for your interest in WearAuthn. The current PR makes ssh-keygen work with WearAuthn. In order for actual SSH authentication to work, #169 is needed to get libfido2 to recognize WearAuthn as a HID device and report it back to SSH. 

I will see whether I can get OpenSSH to support the new pinUvAuthToken once it is no longer hidden behind a flag in libfido2. Then, when credential management has made it to WearAuthn, it should also support the resident key mode of ssh-keygen.

src/fido/types.h
```diff
@@ -16,12 +16,18 @@ extern "C" {
...
+	uint16_t report_in_len;
+	uint16_t report_out_len;
+} fido_dev_io_info_t;
+
+typedef struct fido_dev_io_info *fido_dev_io_open_t(const char *);
+typedef void   fido_dev_io_close_t(struct fido_dev_io_info *);
+typedef int    fido_dev_io_read_t(struct fido_dev_io_info *, unsigned char *, size_t, int);
+typedef int    fido_dev_io_write_t(struct fido_dev_io_info *, const unsigned char *, size_t);
+typedef int    fido_dev_rx_t(struct fido_dev *, uint8_t, unsigned char *, size_t, int);
+typedef int    fido_dev_tx_t(struct fido_dev *, uint8_t, const unsigned char *, size_t);
```
martelletto, line 30: These definitions are part of the public API. Would it be possible to keep them as they are, and cast `void *` to `fido_dev_io_info_t` internally, similar to what hid_openbsd.c and hid_osx.c do?

FabianHenneke: Thanks for catching that, I wasn't aware that these functions were part of the public API until now. Yes, that should be possible, I will force-push a new version once I have it ready.

FabianHenneke: I have amended the first commit to preserve the public API. This is done by including a flag in `fido_dev_t` that allows switching between the native, report size aware io function signature expecting a `fido_dev_io_info_t` and the legacy io functions expecting just a void pointer. If desired, another commit could add a version of `fido_dev_set_io_functions` that is report size aware.

FabianHenneke:
All CI checks pass and I will not make any more changes before the review.

martelletto:
Thanks! I will have a look.

martelletto:
Hi Fabian,

If possible, I would prefer to merge these commits separately. Doing so would allow modifications to be merged and tested more quickly. As such, while I did look at all the commits, I focused my attention on the first one:

- I like the `legacy_io` approach. I suppose we could rename it `fixed_rpt_size` (for clarity), or derive it from `io.open != fido_hid_open` (for obscurity ;-));
- Perhaps we could define a MIN counterpart of `MAX_CTAP_REPORT_LEN` (`CTAP_INIT_HEADER_LEN + 1`?), and call them `CTAP_{MIN,MAX}_REPORT_LEN` (for consistency with other CTAP_* definitions);
- We can probably get rid of the open/close/read/write wrappers by moving `report_{in,out}_len` inside `fido_dev_t` and defining` fido_hid_report_{in,out}_len()` in our HID backends; see 9753f45ae (untested and incomplete code; just a sketch). In this approach, dev.c and hid_*.c are kept separate and do not share types;
- Would it make sense to define `report_{in,out}_len` in `fido_dev_t` as `uint16_t`?

Going forward, would it be OK if I pushed minor fixes (cosmetic nits, mostly) directly to your branch?

Thank you for your work!

-p.

FabianHenneke:
> If possible, I would prefer to merge these commits separately. Doing so would allow modifications to be merged and tested more quickly. 

This makes sense. Would you want me to resubmit as separate PRs our would you split them manually?

> * I like the `legacy_io` approach. I suppose we could rename it `fixed_rpt_size` (for clarity), or derive it from `io.open != fido_hid_open` (for obscurity ;-));

I prefer the 'for clarity' option since it potentially allows custom io functions to take advantage of the support for report length extraction.

> * Perhaps we could define a MIN counterpart of `MAX_CTAP_REPORT_LEN` (`CTAP_INIT_HEADER_LEN + 1`?), and call them `CTAP_{MIN,MAX}_REPORT_LEN` (for consistency with other CTAP_* definitions);

Good idea, this should improve overall consistency.

> * We can probably get rid of the open/close/read/write wrappers by moving `report_{in,out}_len` inside `fido_dev_t` and defining` fido_hid_report_{in,out}_len()` in our HID backends; see [9753f45](https://github.com/Yubico/libfido2/commit/9753f45ae531175b10833ff7d27ab5af5b5d4461) (untested and incomplete code; just a sketch). In this approach, dev.c and hid_*.c are kept separate and do not share types;

You are right, this probably improves separation overall, even though it means that every backend has to define its own context structure. I like your suggestion PR.

> * Would it make sense to define `report_{in,out}_len` in `fido_dev_t` as `uint16_t`?

As in moving it over from 'fido_dev_io_info_t'? Sure!

> Going forward, would it be OK if I pushed minor fixes (cosmetic nits, mostly) directly to your branch?

Go ahead. :-)

martelletto:
> > If possible, I would prefer to merge these commits separately. Doing so would allow modifications to be merged and tested more quickly.
> 
> This makes sense. Would you want me to resubmit as separate PRs our would you split them manually?

I'd prefer PRs, but am fine splitting the commits manually. As you wish.

-p.

FabianHenneke:
> I'd prefer PRs, but am fine splitting the commits manually. As you wish.

I have submitted the first commit, with your suggested changes, as a new PR #172. We could leave this PR open to track the submissions of the individual PRs.

FabianHenneke:
Now that the base PR has been merged it's probably best to close this and deal with the remaining PRs one by one.

PR #169, opened 2020-05-06
Add support for Bluetooth HID devices to hid_linux

FabianHenneke:
Previously, hid_linux.c would fail to extract product and manufacturer
for Bluetooth HID devices since they are not available as individual
sysattrs.

With this commit, the product name is parsed from the uevent sysattr
while the manufacturer, which is not exposed by the kernel, is set to a
descriptive constant ("Bluetooth HID").

I have verified that with this and #168, libfido2 can speak to and detect WearAuthn:
<details>
<summary>Screenshots</summary>

</detail>

### DIFF
--- a/src/hid_linux.c
+++ b/src/hid_linux.c
@@ -8,6 +8,7 @@
 
 #include <sys/ioctl.h>
 #include <linux/hidraw.h>
+#include <linux/input.h>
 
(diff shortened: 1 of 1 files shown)


martelletto:
Thanks! Merging.

PR #172, opened 2020-05-10
Make report lengths configurable in rx/tx/backends

FabianHenneke:
Currently, the lengths of the HID input/output reports are hardcoded as
64 bytes both in the backends and the rx/tx logic in io.c.

This commit makes the report lengths configurable in all these places
without affecting the public API. The report lengths are kept in
`report_{in,out}_len fields` in fido_dev_t and are obtained from the
included backends via fido_hid_report_{in,out}_len() functions. The use
of these functions is gated on a `fixed_rpt_size` flag in fido_dev_t
that could be exposed to custom i/o functions in a future commit.

While the report lengths become configurable with this commit, they are
still kept at the default 64 bytes in all backends. Future commits to
the individual backends will add the logic that extracts the report
lengths from the HID report descriptor.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -411,7 +411,7 @@ subdirs(man)
 
 if(NOT WIN32)
 	if(CMAKE_BUILD_TYPE STREQUAL "Debug")
-		if(NOT MSAN AND NOT LIBFUZZER)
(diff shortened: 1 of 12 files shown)


martelletto:
Merging; thank you!

PR #173, opened 2020-05-10
Read report lengths from descriptor on Linux

FabianHenneke:
Extracts input and output report lengths from the report descriptor
by parsing it manually.

Since the previous implementation did not parse the report lengths out
of the report descriptor, any failure causes a fallback to the default
64 bytes in order to ensure backward compatibility.

This has been tested manually with both a 64 byte and a 62 byte FIDO device.


martelletto:
Thanks! Merging.

PR #177, opened 2020-05-11
Extract credProtect from credman response

FabianHenneke:
The credProtect level of a resident credential is returned under the
key 0x0A (10) of a credential management response. With this commit,
the value is parsed and can be read with fido_cred_prot.

### DIFF
--- a/src/credman.c
+++ b/src/credman.c
@@ -251,9 +251,8 @@ credman_parse_rk(const cbor_item_t *key, const cbor_item_t *val, void *arg)
 		cred->type = cred->attcred.type; /* XXX */
 		return (0);
 	case 10:
-		if (cbor_decode_uint64(val, &prot) < 0 || prot > INT_MAX)
-			return (-1);
-		if (fido_cred_set_prot(cred, (int)prot) < 0)
+		if (cbor_decode_uint64(val, &prot) < 0 || prot > INT_MAX ||
+		    fido_cred_set_prot(cred, (int)prot) != FIDO_OK)
 			return (-1);
 		return (0);
 	default:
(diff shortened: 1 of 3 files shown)


src/credman.c
```diff
@@ -249,6 +250,15 @@ credman_parse_rk(const cbor_item_t *key, const cbor_item_t *val, void *arg)
...
 		cred->type = cred->attcred.type; /* XXX */
 		return (0);
+	case 10:
+		if (cbor_decode_uint64(val, &prot) < 0)
+			return (-1);
+		if (prot != FIDO_CRED_PROT_UV_OPTIONAL &&
+		    prot != FIDO_CRED_PROT_UV_OPTIONAL_WITH_ID &&
+		    prot != FIDO_CRED_PROT_UV_REQUIRED)
+			return (-1);
+		cred->ext.prot = (int)prot;
```
martelletto, line 260: Would it make sense to use fido_cred_set_prot() here?

FabianHenneke: Agreed and force-pushed.

src/credman.c
```diff
@@ -249,6 +250,10 @@ credman_parse_rk(const cbor_item_t *key, const cbor_item_t *val, void *arg)
 			return (-1);
 		cred->type = cred->attcred.type; /* XXX */
 		return (0);
+	case 10:
+		if (cbor_decode_uint64(val, &prot) < 0)
+			return (-1);
+		return (fido_cred_set_prot(cred, (int)prot));
```
FabianHenneke, line 256: Is this conversion fine without range checks?

FabianHenneke: I force-pushed a new version that does mimic the robust check from pin.c and turns every negative return value of fido_cred_set_prot into a -1.

FabianHenneke:
Thanks!

PR #180, opened 2020-05-13
Read report lengths from descriptor on macOS

FabianHenneke:
Extracts input and output report lengths from the report descriptor.

Since the previous implementation rejected all devices with input
report lengths other than 64 bytes, we can reject devices with
invalid report lengths without significantly affecting backward
compatibility - only devices with non-standard HID descriptors that
nonetheless declared a valid input report could be affected negatively.

I confirmed that this builds in the CI, but am unfortunately unable to test it since I lack access to a macOS device.

### DIFF
--- a/src/hid_osx.c
+++ b/src/hid_osx.c
@@ -76,19 +76,51 @@ static bool
 is_fido(IOHIDDeviceRef dev)
 {
 	uint32_t	usage_page;
-	int32_t		report_len;
+	int32_t		report_in_len;
+	int32_t		report_out_len;
 
 	if (get_int32(dev, CFSTR(kIOHIDPrimaryUsagePageKey),
 	    (int32_t *)&usage_page) != 0 || usage_page != 0xf1d0)
 		return (false);
 
 	if (get_int32(dev, CFSTR(kIOHIDMaxInputReportSizeKey),
-	    &report_len) < 0 || report_len != CTAP_MAX_REPORT_LEN) {
-		fido_log_debug("%s: unsupported report len", __func__);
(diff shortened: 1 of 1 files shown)


src/hid_osx.c
```diff
@@ -76,21 +76,53 @@ static bool
...
+		fido_log_debug("%s: unsupported input report len", __func__);
+		return (false);
+	}
+
+	if (get_int32(dev, CFSTR(kIOHIDMaxOutputReportSizeKey),
+	    &report_out_len) < 0 || report_out_len < CTAP_MIN_REPORT_LEN ||
+	    report_out_len > CTAP_MAX_REPORT_LEN) {
+		fido_log_debug("%s: unsupported output report len", __func__);
 		return (false);
 	}
```
martelletto, line 98: Wouldn't it make sense to move these checks to set_report_lengthts(), where we retrieve report_{in,out}_len?

FabianHenneke: My initial intention had been to 

- exclude devices with invalid report sizes (but maintaining backwards compatibility) when no explicit device is specified
- but fall back to 64 bytes when a specifically requested device has invalid report sizes (in dev.c).

But as I only noticed now, this is no longer possible since we got rid of the wrappers - if dev.c unilaterally falls back to 64 bytes, it will potentially be out of sync with the backend report size.

Should I submit a small separate PR that simply fails in `fido_dev_open_tx` when the report sizes reported by the backend are not appropriate? Then, it would make sense for the backends to not do any checks in `set_report_lengths`, but in `is_fido`. Of course we can also change that if you find it confusing.

FabianHenneke: Compare #184, which may also do a better job at explaining my thoughts.

Issue #181, opened 2020-05-13
Linker error on tools in Windows


FabianHenneke:
**What version of libfido2 are you using?**
Current master (https://github.com/Yubico/libfido2/commit/6f08ddf8b9c4fb9ff1d2f0c2f47b7447aca2eb06)

**What operating system are you running?**
Windows

**What application are you using in conjunction with libfido2?**
None

**How does the problem manifest itself?**
After building libfido2, the tools (e.g., fido2-token) are not functional and produce a linker error when run.

**Is the problem reproducible?**
Yes (at least on my machine)

**What are the steps that lead to the problem?**

Running the following in Powershell (with or without admin rights) will produce no output, not even the usual usage information:

```powershell
git clone https://github.com/Yubico/libfido2
cd libfido2
.\windows\build.ps1
.\output\32\bin\fido2-token.exe
.\output\64\bin\fido2-token.exe
```
Running any of the last commands in `cmd` produces a system error saying that `fido2.dll` was not found.

Note that the examples (e.g. `.\build\32\examples\Release\info.exe`) run fine.

**Does the problem happen with different authenticators?**
n/a

martelletto:
IIRC, build.ps1 copies the required dependencies to ${OUTPUT}/pkg. If you try to run fido2-token.exe from ${OUTPUT}/pkg, what happens?

-p.

FabianHenneke:
Thanks, that works perfectly. Guess I started testing with the examples and then wondered why `fido2-token` behaved differently, but I should have just read the script better.

PR #182, opened 2020-05-13
Read report lengths from descriptor on Windows

FabianHenneke:
Extracts input and output report lengths from the report descriptor
using HidP_GetCaps.

Since the previous implementation rejected all devices with report
lengths other than 64 bytes (+ report ID), we can reject devices with
invalid report lengths without affecting backward compatibility.

This has been tested lightly tested with both a 64 byte and a 62 byte FIDO device, further testing will be done after #181 has been resolved.

### DIFF
--- a/src/hid_win.c
+++ b/src/hid_win.c
@@ -48,13 +48,13 @@ is_fido(HANDLE dev)
 	if (caps.OutputReportByteLength > CTAP_MAX_REPORT_LEN + 1 ||
 	    caps.OutputReportByteLength < CTAP_MIN_REPORT_LEN + 1) {
 		fido_log_debug("%s: invalid output report len: %d", __func__,
-		    (int)caps.OutputReportByteLength - 1);
+		    (int)caps.OutputReportByteLength);
 		goto fail;
 	}
 	if (caps.InputReportByteLength > CTAP_MAX_REPORT_LEN + 1 ||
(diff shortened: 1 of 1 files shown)


FabianHenneke:
I have now tested this successfully. Testing showed that previously debug warnings would be emitted for non-FIDO devices, which has now been fixed.

src/hid_win.c
```diff
@@ -42,18 +42,53 @@ is_fido(HANDLE dev)
...
+		fido_log_debug("%s: invalid output report len: %d", __func__,
+		    caps.OutputReportByteLength - 1);
+		goto fail;
+	}
+	if (caps.InputReportByteLength - 1 > CTAP_MAX_REPORT_LEN ||
+	    caps.InputReportByteLength - 1 < CTAP_MIN_REPORT_LEN) {
+		fido_log_debug("%s: invalid input report len: %d", __func__,
+		    caps.InputReportByteLength - 1);
 		goto fail;
 	}
```
martelletto, line 59: silly suggestion: could we turn the subtraction into an addition on the right side of the inequation, so we don't depend on the type of OutputReportByteLength/InputReportByteLength to handle the < 1 case?

FabianHenneke: Should be done. I also added an explicit cast to `int` to the debug statements.

martelletto:
Thanks! Merging.

PR #184, opened 2020-05-14
Fail in fido_dev_open_tx on invalid report lengths

FabianHenneke:
Since fido_dev_t and the backends separately maintain the report
lengths, fido_dev_open_tx cannot unilaterally fall back to 64 bytes and
has to fail if it encounters an invalid report length.

With this change, the backends no longer have to perform any checks on
report lengths extracted from the report descriptor. They may however
choose to do so to help them decide whether or not a device is a valid
FIDO device.

### DIFF
--- a/src/dev.c
+++ b/src/dev.c
@@ -158,16 +158,18 @@ fido_dev_open_tx(fido_dev_t *dev, const char *path)
 	    dev->report_in_len > CTAP_MAX_REPORT_LEN) {
 		fido_log_debug("%s: invalid report_in_len %hu", __func__,
 		    dev->report_in_len);
-		/* Fall back to default USB transfer report length. */
-		dev->report_in_len = CTAP_MAX_REPORT_LEN;
+		dev->io.close(dev->io_handle);
+		dev->io_handle = NULL;
+		return (FIDO_ERR_RX);
 	}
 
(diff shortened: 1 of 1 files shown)


src/dev.c
```diff
@@ -158,16 +158,14 @@ fido_dev_open_tx(fido_dev_t *dev, const char *path)
 	    dev->report_in_len > CTAP_MAX_REPORT_LEN) {
 		fido_log_debug("%s: invalid report_in_len %hu", __func__,
 		    dev->report_in_len);
-		/* Fall back to default USB transfer report length. */
-		dev->report_in_len = CTAP_MAX_REPORT_LEN;
+		return (FIDO_ERR_INTERNAL);
```
martelletto, line 161: Makes sense. Given we have opened the device and assigned dev->io_handle at that point, we need to

```c
dev->io.close(dev->io_handle);
dev->io_handle = NULL;
```

before returning.

FabianHenneke: Thanks! I fixed that and also changed the returned error code to distinguish between issues with the in or out report - I found `FIDO_ERR_TX` and `FIDO_ERR_RX` appropriate since these would have been the ultimately returned error codes before we made libfido2 report size aware.

PR #189, opened 2020-07-01
Add note about `pkg-config` to the readme

jesse-savary:
pkg-config is needed for the calls to `pkg_search_module` in https://github.com/Yubico/libfido2/blob/master/CMakeLists.txt

See https://cmake.org/cmake/help/latest/module/FindPkgConfig.html

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -60,12 +60,11 @@ Or from source, on UNIX-like systems:
   $ make -C build
   $ sudo make -C build install
 
-Depending on the platform, the PKG_CONFIG_PATH environment variable may need to
-be set.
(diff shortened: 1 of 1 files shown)


martelletto:
Hi,

Thank you for the PR. For clarity, I suggest moving the reference to `pkg-config` to the previous line, where we talk about `PKG_CONFIG_PATH`.

-p.

jesse-savary:
Made an attempt at rewording things, not sure if that sentence is clear enough. I hope `PKG_CONFIG_PATH` implies that it needs to be set to the location of the user's current `pkg-config` installation.

martelletto:
Thank you! Merged.

PR #197, opened 2020-08-05
Add tool detecting unprotected resident credentials

FabianHenneke:
Adds `fido2-unprot` which scans attached security keys for resident
credentials that do not require user verification to be enumerated.

If a security key does not support Credential Management, the script
only checks whether there is a unprotected resident SSH credential
created with default settings. Detecting the latter is the primary
usecase of this script.


martelletto:
thanks; merging

PR #198, opened 2020-08-06
Improve fido2-unprot compatibility

FabianHenneke:
Improves fido2-unprot in the following ways, following https://github.com/Yubico/libfido2/commit/734b0fed05a81dc6d01e712776cf1dd26d5a8826#r41226118:

* Use global fido2-* binaries (override with FIDO_TOOLS_PREFIX)
* Exit if it is not running on Linux
* Exit if libfido2 is not installed or not recent enough
* Exit if an error occurs

### DIFF
--- a/tools/fido2-unprot.sh
+++ b/tools/fido2-unprot.sh
@@ -4,8 +4,28 @@
 # Use of this source code is governed by a BSD-style
 # license that can be found in the LICENSE file.
 
-PREFIX=~/git/libfido2/build/tools/
-TOKEN_OUTPUT=$(${PREFIX}fido2-token -L)
+
+if [ $(uname) != "Linux" ] ; then
+   echo "Can only run on Linux"
+   exit 1
+fi
+
+TOKEN_VERSION=$(${FIDO_TOOLS_PREFIX}fido2-token -V 2>&1)
+if [ $? -ne 0 ] ; then
+    echo "Please install libfido2 1.5.0 or higher"
+    exit
+fi
+
+TOKEN_VERSION_MAJOR=$(echo "$TOKEN_VERSION" | cut -d. -f1)
+TOKEN_VERSION_MINOR=$(echo "$TOKEN_VERSION" | cut -d. -f2)
+if [ $TOKEN_VERSION_MAJOR -eq 0 -o $TOKEN_VERSION_MAJOR -eq 1 -a $TOKEN_VERSION_MINOR -lt 5 ] ; then
+    echo "Please install libfido2 1.5.0 or higher (current version: $TOKEN_VERSION)"
+    exit 1
+fi
+
+set -e
+
+TOKEN_OUTPUT=$(${FIDO_TOOLS_PREFIX}fido2-token -L)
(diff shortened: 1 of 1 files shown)


nevun:
The failed CIFuzz build step is known: https://github.com/google/oss-fuzz/issues/4262

tools/fido2-unprot.sh
```diff
@@ -15,37 +35,37 @@ do
...
             fi
         done
-        if test $FOUND -eq 0 ; then
+        if [ $FOUND -eq 0 ] ; then
             echo "No unprotected credentials on $DEV_PRETTY"
         fi
     else
         echo "$DEV_PRETTY cannot enumerate credentials"
         echo "Discovering unprotected SSH credentials only..."
         STUB_HASH=$(echo -n "" | sha256sum)
```
nevun, line 66: With that I get: 
fido2-assert: fido_dev_get_assert: FIDO_ERR_INVALID_LENGTH

You probably want:
STUB_HASH=$(echo -n "" | openssl sha256 -binary | base64)

Then I get: 
fido2-assert: fido_dev_get_assert: FIDO_ERR_NO_CREDENTIALS

FabianHenneke: I changed this to your recommendation. Both work for me on Linux, does this change mean that this now works for you on mac OS?

I am surprised that you are seeing the error message since stderr should have been redirected to `/dev/null`.

nevun: It does not work yet on MacOS and would require several other changes for that.

Could it be that it worked for you because you never tested it with an unprotected SSH credential? 

FIDO_ERR_NO_CREDENTIALS and FIDO_ERR_INVALID_LENGTH exit codes both results in going through the else clause "No unprotected SSH credentials" by your script as it is now.

nevun:
> Would you mind changing the commit message to:
> 
> "base64 encode the binary sha256 hash "
> 
> or something similar which more correctly reflects the change?
> 
> Thanks!

never mind, I'll just squash it when merging.

nevun:
Rebased, squashed and merged in https://github.com/Yubico/libfido2/commit/42c6957fece8d2a80815603256760eb431fd297f

tools/fido2-unprot.sh
FabianHenneke: Thanks for changing the commit message.

> Could it be that it worked for you because you never tested it with an unprotected SSH credential?
> 
> FIDO_ERR_NO_CREDENTIALS and FIDO_ERR_INVALID_LENGTH exit codes both results in going through the else clause "No unprotected SSH credentials" by your script as it is now.

I just tested this with a device without support for credential management and with no credentials, which should go through the correct branch. This is when `fido2-assert` should print `fido2-assert: fido_dev_get_assert: FIDO_ERR_NO_CREDENTIALS` to `stderr`, given that it fails to find a credential for the RP `ssh:`, but this is supressed by the redirect of `stderr` to `/dev/null`. It does print the message if I remove the `2> /dev/null` from `fido2-assert` in that case.

PR #204, opened 2020-08-26
udev/fidodevs: Added PID to fidodevs for the MIRKey

kobusgrobler:
Added support for the MIRKey.

Thanks


martelletto:
Thank you!

PR #205, opened 2020-08-27
make master build on OpenBSD current

nevun:
No idea if this is the right thing to do but now it builds and the other builds seem happy too :)


martelletto:
Thank you!

PR #213, opened 2020-09-09
Ubuntu 20 can use the same instructions as 18,16?

gabriel:
The Ubuntu package is still at 1.3.1 here:
https://packages.ubuntu.com/source/focal/libfido2


martelletto:
Hi @gabriel, thank you for the PR! I added a blurb in 1fb0df7 to that effect, with the understanding that pointing to official packages by default (where available, and even if outdated) is recommended practice.

-p.

PR #224, opened 2020-10-02
fuzz: revert 5032dd78; oss-fuzz instances don't provide getrandom

nevun:


martelletto:
looks good to me. if only i could remember what i did 11 months ago... apologies for the breakage.

PR #235, opened 2020-11-02
cbor: recursively check cbor encoding for outgoing messages

LDVG:
In debug builds, this validates the CTAP2 canonical CBOR encoding form
recursively for all messages that are built using `cbor_build_frame()`.
Checks include that all items are of a definite length, that map
keys are sorted correctly, and that integers are encoded as small as
possible.


LDVG:
Closing after internal benefit vs. complexity discussion :slightly_smiling_face:

PR #240, opened 2020-11-06
udev: add Nitrokey 3C NFC

szszszsz:
Hi!

With this PR I would like to add Udev rule for Nitrokey 3C NFC.

cc @Nitrokey


martelletto:
Thank you!

PR #243, opened 2020-11-20
Add a NetBSD back end.

riastradh:
Previously we just used the OpenBSD one, but that no longer works
after 1.5.0 since NetBSD doesn't expose /dev/fido/* nodes.


martelletto:
Thank you!

PR #245, opened 2020-12-04
Add/Update Trustkey T110, T120, G310, G320

7tobias:
According to https://www.trustkeysolutions.com/support/ two new products
T110 and T120 are added. The formerly eWBM G310, G320 are now under the
Trustkey brand.

### DIFF
--- a/udev/70-u2f.rules
+++ b/udev/70-u2f.rules
@@ -184,22 +184,22 @@ KERNEL=="hidraw*", SUBSYSTEM=="hidraw", ATTRS{idVendor}=="2c97", ATTRS{idProduct
 # Hypersecu HyperFIDO by Hypersecu Information Systems, Inc.
 KERNEL=="hidraw*", SUBSYSTEM=="hidraw", ATTRS{idVendor}=="2ccf", ATTRS{idProduct}=="0880", TAG+="uaccess", GROUP="plugdev", MODE="0660"
 
-# TrustKey Solutions FIDO2 TrustKey G310 by eWBM Co., Ltd. (Trustkey Solutions)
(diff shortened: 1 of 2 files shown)


martelletto:
Thanks. Some remarks:

- according to https://www.usb.org/sites/default/files/vendor_ids081020_1.pdf, the vendor ID `0x311f` belongs to `eWBM Co., Ltd.` I suggest keeping that name;
- from the link provided, the "K" in TrustKey is uppercase;
- the `vendor` and `product` sections of `fidodevs` need to be sorted: https://github.com/Yubico/libfido2/blob/master/udev/check.sh

7tobias:
> Thanks. Some remarks:
> 
> * according to https://www.usb.org/sites/default/files/vendor_ids081020_1.pdf, the vendor ID `0x311f` belongs to `eWBM Co., Ltd.` I suggest keeping that name;
I changed it back to eWBM - Anyway in context of Fido eWBM is now the hardware producer and Trustkey Solutions is now vendor of the sticks (software, firmware, distribution, support, logo on packaging and sticks etc) even ewbm.com is now just a forward to trustkeysolutions.com
> * from the link provided, the "K" in TrustKey is uppercase;
you are right - i changed it
> * the `vendor` and `product` sections of `fidodevs` need to be sorted: https://github.com/Yubico/libfido2/blob/master/udev/check.sh
i sorted it, hope it's fine - the check script is not working in my place.

martelletto:
Thanks for the additional context. I have kept `eWBM Co., Ltd.` since that is what is listed on www.usb.org. When they publish a new list of vendor IDs, we will update the entries accordingly.

PR #248, opened 2020-12-08
pin: add fido_dev_get_uv_retry_count()

LDVG:
This retrieves the number of built-in UV attempts before built-in UV is
disabled on the device.


martelletto:
Thank you!

Issue #251, opened 2020-12-21
Atomic signal masks for reliably interruptible device I/O


riastradh:
**Describe the feature you would like.**
I would like to transact asynchronously with the first of any (reasonable) number of authenticator devices in the system, like https://github.com/mozilla/authenticator-rs does.  One way to do this would be to completely redesign the API to be asynchronous and event-driven (and it would also be nice to wait for the user to insert the device, by waiting for udev/drvctl/&c. events).  But that's obviously a lot of work.

Short of that, querying the devices in separate pthreads will serve as an acceptable alternative: according to https://github.com/Yubico/libfido2/issues/43, libfido2 generally allows operations on different objects in different threads as long as the threads themselves serialize the access, and there should usually be few enough devices that a profusion of threads is not an issue.  Rough sketch (omitting details of locking, error checking, and so on):

```
/* Main thread */
pthread_create(worker, devinfo1);
pthread_create(worker, devinfo2);
...
while (!done) {
	if (pthread_cond_timedwait(...) == ETIMEDOUT)
		break;
}

/* Worker N */
devN = fido_dev_new();
fido_dev_open(devN, devinfoN);
fido_dev_make_cred(devN, credN);
done = 1;
pthread_cond_signal(...);
```

But there's a snag.  Once the _first_ device has answered a transaction such as `fido_dev_make_cred`, the other transactions and threads need to be interrupted so they don't continue to hang waiting for input that's not coming and would be discarded anyway.  Two obvious approaches are:

1. Set up a null signal handler, and have the main thread deliver SIGUSR1 to all the worker threads that haven't completed yet, so that (e.g.) `poll` in `fido_hid_unix_wait` will return early with EINTR.
2. When a worker thread completes, have the main thread issue `fido_dev_cancel` for all devices that haven't returned yet, so that (e.g.) the CBOR authenticatorMakeCredential transaction will abort with error.

However, both of these approaches necessarily have race conditions, because there's no way for the main thread to know exactly when the CTAP transaction has actually begun: deliver a signal too early, and it won't interrupt the `poll` inside libfido2 because `poll` hasn't started yet; cancel a transaction too early, and it won't interrupt the pending transaction because the transaction isn't pending yet.  Either way, the worker thread will hang when it should respond promptly to interruption.

[...]

PR #252, opened 2020-12-22
Pass signal mask down into I/O functions.

riastradh:
This is a draft proof-of-concept to pass the signal mask down into waits for I/O -- see https://github.com/Yubico/libfido2/issues/251 for details.

Caveats in this draft:

 - This changes the API and ABI of fido_dev_set_io_functions and fido_dev_set_transport_functions -- the functions now take a signal mask to set while waiting for I/O.

 - Formerly fido_hid_unix_wait would loop on EINTR, but that renders the whole exercise moot -- the point is to abort and pass the buck back to the caller.  Looping on EINTR also only happened when waiting in poll via fido_hid_unix_wait, not in read or write; it's not clear to me why this loop was added.

### DIFF
--- a/regress/assert.c
+++ b/regress/assert.c
@@ -52,9 +52,10 @@ static const unsigned char sig[72] = {
 };
 
 static void *
-dummy_open(const char *path)
+dummy_open(const char *path, const sigset_t *sigmask)
 {
 	(void)path;
+	(void)sigmask;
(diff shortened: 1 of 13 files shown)


riastradh:
superseded by https://github.com/Yubico/libfido2/pull/254

PR #254, opened 2020-12-22
New function fido_dev_set_sigmask.

riastradh:
This sets the signal mask to be in effect atomically during all waits
for I/O.  Atomically setting the signal mask enables reliable
interruption of I/O waits, e.g. so multiple threads can query multiple
different devices, and whichever one gets an answer first (perhaps
because the user tapped a device) can reliably send signals to
interrupt all the other ones.

fix https://github.com/Yubico/libfido2/issues/251

(An asynchronous event-driven push-oriented state machine might be
better, but that would require substantially restructuring the entire
libfido2 API and implementation.)

CAVEATS:

- This changes the ABI of fido_dev_set_io_functions: fido_dev_io_t
  now has a new member, set_sigmask, for setting the signal mask.
  There's no API change (in contrast to the earlier version,
  <https://github.com/Yubico/libfido2/pull/252>) -- you just can't
  use fido_dev_set_sigmask if you do fido_dev_set_io_functions with a
  null set_sigmask callback -- but clients will have to be
  recompiled.

- Formerly fido_hid_unix_wait would loop on EINTR, but that renders
  the whole exercise moot -- the point is to abort and pass the buck
  back to the caller.  Looping on EINTR also only happened when
  waiting in poll via fido_hid_unix_wait, not in read or write; it's
  not clear to me why this loop was added.

- I scaled back a little from the earlier version,
  <https://github.com/Yubico/libfido2/pull/252> -- that patch issues
  interruptible ppoll before write too, not just before read.  But
  realistically, writes are unlikely to block anyway because the
  caller should have exclusive access to the device: there should be
  no queue of pending outgoing interrupt transfers we have to wait
  for anyway, and ppoll can't know how long the write will block
  anyway.  (It would be nice if the write system call itself could
  take a signal mask, but it doesn't, nor do any of the standard
  variants pwrite/writev/pwritev.)


riastradh:
superseded by https://github.com/Yubico/libfido2/pull/255 (_POSIX_C_SOURCE cleanup discarded)

Issue #256, opened 2020-12-27
fido_assert_set_hmac_secret


riastradh:
**Describe the feature you would like.**
A function `fido_assert_set_hmac_secret(fido_assert_t *assert, size_t idx, const unsigned char *ptr, size_t len)` to set the hmac-secret, for later retrieval with `fido_assert_hmac_secret_ptr` and `fido_assert_hmac_secret_len`, much like the existing `fido_assert_set_authdata_raw` and others.  Although not useful for applications to use directly, this would make it easier to test functionality that depends on the hmac-secret of an assertion response downstream, without having to mock up the entire interaction including the ECDH key agreement between authenticator and host.

**Describe the impact to existing functionality.**
none

**Describe how you are willing to help.**
Happy to draft a patch if you like.

PR #257, opened 2020-12-27
New function fido_assert_set_hmac_secret.

riastradh:
Useful for testing applications that use the hmac-secret.

fix https://github.com/Yubico/libfido2/issues/256


martelletto:
Thank you!

PR #265, opened 2021-01-15
udev: Add CanoKey and Update to correct vendor

ZenithalHourlyRate:
Hi, with this PR a udev rule for a new fido device `Canokey` is added. Please check @Canokeys for more info!

Also, the vendor is changed from `Flirc` to `Clay Logic` as `Clay Logic` is the correct holder of the vendorID.


martelletto:
Thank you!

PR #272, opened 2021-02-02
hmac-secret: fix regression for case where no pin is supplied

LDVG:


src/pin.c
```diff
@@ -691,8 +691,14 @@ cbor_add_uv_params(fido_dev_t *dev, uint8_t cmd, const fido_blob_t *hmac_data,
...
 {
-	return (pin ||
-	    (fido_dev_supports_permissions(dev) && fido_dev_has_uv(dev)));
+	if (pin != NULL)
+		return (true);
+	if (fido_dev_supports_permissions(dev)) {
+		if (uv != FIDO_OPT_OMIT)
+			return (uv == FIDO_OPT_TRUE);
+		return (fido_dev_has_uv(dev));
+	}
```
martelletto, line 702: can you please add a newline here?

LDVG: fixed!

PR #284, opened 2021-03-05
miscellaneous improvements in fido2-token

martelletto:


LDVG (approved):
This looks all good to me :)

tools/util.c
```diff
@@ -584,3 +570,9 @@ plural(size_t x)
 {
 	return x == 1 ? "" : "s";
 }
+
+int
+should_retry_with_pin(int r)
+{
+	return r == FIDO_ERR_PIN_REQUIRED || r == FIDO_ERR_UV_INVALID;
```
LDVG, line 577: for authenticators that do not support `clientPin` this will prompt the user for a non-existing pin if built-in UV fails, right?

kongeo (approved):
looks good @martelletto

PR #288, opened 2021-03-09
disable nfc support in preparation for release

martelletto:


kongeo (approved):

martelletto:
Yes, this is the experimental Linux NFC support. It was disabled in the sense that it won't be available out of the box, but through a compile-time option. There are some rough edges we need to address (happy to elaborate if there is interest). Further, we do not want to surprise anyone when they update a package and applications suddenly start talking NFC - which is the case with the code as it stands. We are also not aware of anyone using the feature.

rileyg98:
I would be interested in the 'rough edges' - I am currently exploring the use of this library with NFC support for a NFC-only implementation of FIDO2 (currently being developed by myself), for further use in libpam-u2f, and would be happy to address these rough edges to allow the possibility for "official" NFC support - currently, it uninstalls a few "big" parts of Ubuntu when uninstalling libfido2 provided by Ubuntu, to replace it with this version. We would, as a result, be using this feature - happy to discuss this in greater depth.

martelletto:
@rileyg98 Great to hear! The plumbing is there (in HEAD, or in 1.7.0 built with `-DNFC_LINUX`), and the functionality should work. The rough edges we need to address are:

- Is the device path (as implemented) acceptable, or would a opaque, umbrella path (e.g. `linux://nfc`) be preferable?
- We currently block on a NFC reader until a device is detected. Is this behaviour desired?
- Should we try to detect NFC devices out of the box?
- What is the impact to applications using libfido2?

hramrach:
There are sites that recommend to use an USB token as main and wireless as backup (I think it's bluetooth but it applies equally to NFC).

From use PoV there should be no difference between different connection path of tokens, the connection paths should be interchangeable, and the connection path used should be autodetected.

From that 

 - NFC devices should be detected automatically when NFC is available
 - NFC detection should not block - eg. polling the reader for devices a few times per second might work well enough
 - NFC support should not prevent the user from using an USB token

martelletto:
This will be revisited in 1.9.0. Thank you for the input.

PR #289, opened 2021-03-09
udev: Add alternative ID for GoTrust Idem Key

hramrach:
https://www.gotrustid.com/faqs-of-idem-key
https://download.gotrustid.com/Idem/IdemKey/70-u2f.rules

Signed-off-by: Michal Suchanek <msuchanek@suse.de>

Only based on the rules file - not tested.


martelletto:
Thank you!

hramrach:
Verified current device has the new USB ID

Bus 003 Device 112: ID 32a3:3201 GoTrust Idem Key

PR #302, opened 2021-03-17
introduce FIDO_ERR_UNAUTHORIZED_PERM; handle it in should_retry_with_pin()

kongeo:
CTAP2_1 supports a uvAcfg option that flags whether the token allows the
acfg permission for getting a token with uv or not. This commit
showcases how that option can be used to obtain a token for the config
command.


tools/config.c
```diff
@@ -92,21 +92,26 @@ config_pin_minlen(char *path, const char *pinlen)
...
 	dev = open_dev(path);
 	if ((len = base10(pinlen)) < 0 || len > 63) {
 		warnx("%s: len > 63", __func__);
 		goto out;
 	}
-	if ((r = fido_dev_set_pin_minlen(dev, (size_t)len, NULL)) != FIDO_OK &&
-	    should_retry_with_pin(dev, r)) {
-		if ((pin = get_pin(path)) == NULL)
+	if ((get_devopt(dev, "uvAcfg", &v) == 0) && (v > 0)) {
+		if ((r = fido_dev_set_pin_minlen(dev, (size_t)len, NULL)) != FIDO_OK &&
```
LDVG, line 103: If the device supports uvAcfg, wont this first set the min length using built-in uv, then prompt the user for a PIN and set the min length again?

tools/config.c
```diff
@@ -125,17 +130,22 @@ config_force_pin_change(char *path)
...
 	char *pin = NULL;
-	int r, ok = 1;
+	int v, r, ok = 1;
 
 	dev = open_dev(path);
-	if ((r = fido_dev_force_pin_change(dev, NULL)) != FIDO_OK &&
-	    should_retry_with_pin(dev, r)) {
-		if ((pin = get_pin(path)) == NULL)
+	if ((get_devopt(dev, "uvAcfg", &v) == 0) && (v > 0)) {
+		if ((r = fido_dev_force_pin_change(dev, NULL)) != FIDO_OK &&
```
LDVG, line 137: dito above

tools/config.c
kongeo: ohh right, that scenario :)

martelletto (approved):

LDVG (approved):

PR #303, opened 2021-03-18
fuzz: wrap a few more libcrypto symbols

martelletto:
N.B. please double-check that the return values listed in wrap.c are correct.


LDVG (approved):
This looks correct to me :)

kongeo (approved):

PR #304, opened 2021-03-19
address a few inconsistencies in what we include, and how we include it

martelletto:


LDVG (approved):
Looks reasonable to me!

kongeo (approved):

PR #305, opened 2021-03-19
credman: allow requests without a pin/uv token through

martelletto:
so that the authenticator has a chance to reply with `FIDO_ERR_PIN_REQUIRED`, as per spec. this is a sensible thing to do on its own, but also required for `fido2-token` to work as expected with PIN authenticaton after https://github.com/Yubico/libfido2/commit/6c1d0886e2d8451b857505744b5b83583924c38b.


LDVG (approved):
Makes sense to me!

kongeo (approved):

PR #306, opened 2021-03-24
assert: fix use of fido_dev_can_get_uv_token()

martelletto:
calling `fido_dev_can_get_uv_token()` from the module's `tx()` function is typically ok, but assertions are a special case: we obtain a ecdh shared secret in the very beginning, and hold on to it while making multiple trips to the authenticator, so that we can gather all the assertions and decrypt any hmac secrets. we should therefore move the call to `fido_dev_can_get_uv_token()` from `tx()` to the outer `fido_dev_get_assert()`, where the call to `fido_do_ecdh()` takes place.

### DIFF
--- a/src/assert.c
+++ b/src/assert.c
@@ -132,7 +132,7 @@ fido_dev_get_assert_tx(fido_dev_t *dev, fido_assert_t *assert,
 		}
 
 	/* user verification */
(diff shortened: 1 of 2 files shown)


LDVG (approved):

kongeo (approved):
lgtm!

Issue #316, opened 2021-03-31
Resulting Shared Library in wrong destination


xgcssch:
**What version of libfido2 are you using?**
1.7

**How does the problem manifest itself?**
When building fido2 as a shared library, the resulting Object is not placed in the correct directory according

https://www.gnu.org/prep/standards/html_node/Directory-Variables.html
https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html
https://cmake.org/cmake/help/latest/command/install.html

This is due to specifying

```
RUNTIME DESTINATION ${CMAKE_INSTALL_LIBDIR})
```

in line 123 of file `src/CMakeLists.txt`.

It should be changed to 

```
RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR})
```

martelletto:
Fixed in d3ed2bfc1172074f81ce04973f33db532c4ec141. Thank you!

PR #317, opened 2021-03-31
Changed installation directory of DLLs on Windows

xgcssch:
As stated in 
https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html or https://cmake.org/cmake/help/latest/command/install.html
the target for RUNTIME DESTINATION should be CMAKE_INSTALL_BINDIR and not CMAKE_INSTALL_LIBDIR

This is a proposed fix for #316


martelletto:
The changes are fine, but the commit message of f28aba1 could use some rephrasing: what is effectively being changed is the installation directory of DLLs on Windows, and nothing else. The documentation on cmake.org borrows the term "GNU" and uses it in a rather loose way to describe installation paths that are common not only in GNU, but in UNIX-like environments in general, and later extrapolates the term to include non-UNIX environments as well. As far as UNIX-like environments and paths are concerned, commit f28aba1  is a no-op. Could I bother you to please amend the commit message? Thank you!

xgcssch:
Your were right: My wording violated KISS ;-)

martelletto:
Thank you. I've pushed a new branch to get rid of the merge commit.

PR #330, opened 2021-05-21
cmake: remove code duplication

martelletto:
populate TARGET_LIBRARIES once and use it in the definition of the static and shared lib targets, instead of duplicating the contents of TARGET_LIBRARIES in both places.


LDVG (approved):
Looks good! :+1:

PR #331, opened 2021-05-21
pkg-config: revert 4b9ff83

martelletto:
While there may be a point in defining 'Requires.private', it is ultimately the territory of package maintainers and wasn't the commit's intention; 'Libs.private' was intended.

Due to CMake's architecture and the concept of a separate "generator", expanding 'Libs.private' from within CMakeLists.txt isn't trivial. While there are some acrobatics we could do, they are likely to produce innacurate results.


martelletto:
4b9ff83 was introduced in PR #293.

LDVG (approved):

PR #336, opened 2021-06-15
Initial support for Windows's native WebAuthn API

### LINKED ISSUE #34: windows: abstraction layer for system's native API
I have a problem with the libfido2 detecting any attached key on my Surface Book 2. I have tried multiple keys and using webauth demos or the YubicoManager they all successfully detect the key, but the number of devices from the `fido_dev_info_manifest` call is always 0. Running the same code on other laptops/desktops works with the library detecting multiple keys (`fido_dev_info_manifest` returning > 0 devices). 

The drivers for the keys always installs correctly and Windows recognize the key as a security device etc. I have tried keys from Yubico and some keys from other vendors.

The Windows version I have installed is 1903 (release preview) but I have tried attaching the key using a Virtual Machine with Windows 7 and earlier Windows 10 versions and still the same problem.

```
fido_dev_t *device = NULL;
fido_dev_info_t *deviceInfoList;
size_t ndevs = 0;
fido_init(0);
if ((deviceInfoList = fido_dev_info_new(64)) && fido_dev_info_manifest(deviceInfoList, 64, &ndevs) == FIDO_OK)
```

This code will return `FIDO_OK` but with `ndevs` set to 0 with a key attached to the computer.

martelletto:
This PR gives applications the opportunity to talk to FIDO devices through Windows's native WebAuthn API, which gets abstracted by libfido2 as a FIDO device with path `windows://hello`. Applications willing to make use of this abstraction will need to:

- use `fido_cred_set_clientdata()` and `fido_assert_set_clientdata()` instead of `fido_cred_set_clientdata_hash()` and `fido_assert_set_clientdata_hash()`;
- not prompt for the PIN themselves, but instead pass a non-NULL PIN to `fido_dev_make_cred()` or `fido_dev_get_assert()` (or set UV in the corresponding credential/assertion), so that Windows's system API prompts for the PIN. 

Please note that the only extension supported at the moment is `credProtect`.

Fixes #34.


kongeo (approved):
Nice! LGTM :)

LDVG (approved):
Nice work! :ok_hand:

PR #351, opened 2021-07-06
io: discard frames with wrong channel identifier

LDVG:
If multiple applications are accessing a single device both would error
out since they'd intercept frames with the wrong channel identifier
(cid). Instead, discard all frames with the wrong cid and loop until we
receive the response to our message.


martelletto:
LGTM.

martelletto (approved):

kongeo (approved):

PR #354, opened 2021-07-12
fuzz_hid: fuzz udev bits

martelletto:
wrap enough of libudev to allow the fuzzing of `fido_hid_manifest()` and `fido_nfc_manifest()` on linux. short of using libudev directly (which is impractical due to its behaviour being filesystem bound), this approach is probably the best we can do.


LDVG (approved):
Nice!

kongeo (approved):

PR #357, opened 2021-07-28
Replace SHA256_* with EVP_* equivalents

martelletto:
SHA256_{Init,Update,Final} are marked as deprecated in OpenSSL 3.0. Given that alternatives exist in OpenSSL 1.1 and LibreSSL, switch to them as a first step towards OpenSSL 3.0 integration. Joint work with Dmitry Belyavskiy (@beldmit).

### DIFF
--- a/fuzz/wrap.c
+++ b/fuzz/wrap.c
@@ -136,30 +136,6 @@ WRAP(int,
 	1
 )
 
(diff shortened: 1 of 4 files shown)


kongeo (approved):
LGTM!

PR #363, opened 2021-08-03
cred: rework verify_sig

martelletto:
Rework verify_sig (used by fido_cred_verify) to accept key types other than ES256, while making the code OpenSSL 3.0 compatible.

Diff from Dmitry Belyavskiy (@beldmit).


kongeo (approved):
LGTM!

PR #395, opened 2021-09-13
pinUvAuthProtocol parameter in HMAC-Secret assertion

niryanowsky:
In 1.7.0 cbor_encode_hmac_secret_param add a new fourth parameter ( pinUvAuthProtocol ) to HMAC Secret Extension as part of GetAssertion.
This parameter was added (even thought optional) to hmac-secret authenticatorGetAssertion request in FIDO 2.1, and not in FIDO 2.0 (regardless the existence of pin protocol).
So, some kind of version checking should apply before adding the new parameter.
This parameter make feitian fido2 authenticator to return FIDO_ERR_UNSUPPORTED_EXTENSION when try the get assertion with FIDO_EXT_HMAC_SECRET.

### DIFF
--- a/src/cbor.c
+++ b/src/cbor.c
@@ -817,13 +817,20 @@ cbor_encode_hmac_secret_param(const fido_dev_t *dev, cbor_item_t *item,
 	/* XXX not pin, but salt */
 	if ((argv[0] = es256_pk_encode(pk, 1)) == NULL ||
 	    (argv[1] = fido_blob_encode(enc)) == NULL ||
-	    (argv[2] = cbor_encode_pin_auth(dev, ecdh, enc)) == NULL ||
-	    (argv[3] = cbor_encode_pin_opt(dev)) == NULL) {
+	    (argv[2] = cbor_encode_pin_auth(dev, ecdh, enc)) == NULL ) {
 		fido_log_debug("%s: cbor encode", __func__);
 		r = FIDO_ERR_INTERNAL;
 		goto fail;
 	}
 
+	if (fido_dev_supports_pin_protocol_in_hmac_assert(dev)) {
+            if ((argv[3] = cbor_encode_pin_opt(dev)) == NULL) {
+                fido_log_debug("%s: cbor encode", __func__);
+                r = FIDO_ERR_INTERNAL;
+                goto fail;
+            }
+    	}
+	
 	if ((param = cbor_flatten_vector(argv, nitems(argv))) == NULL) {
 		fido_log_debug("%s: cbor_flatten_vector", __func__);
 		r = FIDO_ERR_INTERNAL;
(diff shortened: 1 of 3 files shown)


martelletto:
Thank you for the PR.

[FIDO 2.0 says](https://fidoalliance.org/specs/fido-v2.0-ps-20190130/fido-client-to-authenticator-protocol-v2.0-ps-20190130.html#message-encoding):

> If map keys are present that an implementation does not understand, they MUST be ignored. Note that this enables additional fields to be used as new features are added without breaking existing implementations.

which the authenticator you are using does not seem to observe.

[FIDO 2.1 says](https://fidoalliance.org/specs/fido-v2.1-ps-20210615/fido-client-to-authenticator-protocol-v2.1-ps-20210615.html#sctn-hmac-secret-extension):

> pinUvAuthProtocol(0x04): (optional) as selected when getting the shared secret. CTAP2.1 platforms MUST include this parameter if the value of pinUvAuthProtocol is not 1.

which makes me think the following to be a smaller, more localised workaround:

```diff
diff --git a/src/cbor.c b/src/cbor.c
index 6cd0087..7935e50 100644
--- a/src/cbor.c
+++ b/src/cbor.c
@@ -788,6 +788,7 @@ cbor_encode_hmac_secret_param(const fido_dev_t *dev, cbor_item_t *item,
        cbor_item_t             *argv[4];
        struct cbor_pair         pair;
        fido_blob_t             *enc = NULL;
+       uint8_t                  prot;
        int                      r;
 
        memset(argv, 0, sizeof(argv));
@@ -814,11 +815,17 @@ cbor_encode_hmac_secret_param(const fido_dev_t *dev, cbor_item_t *item,
                goto fail;
        }
 
+       if ((prot = fido_dev_get_pin_protocol(dev)) == 0) {
+               fido_log_debug("%s: fido_dev_get_pin_protocol", __func__);
+               r = FIDO_ERR_INTERNAL;
+               goto fail;
+       }
+
        /* XXX not pin, but salt */
        if ((argv[0] = es256_pk_encode(pk, 1)) == NULL ||
            (argv[1] = fido_blob_encode(enc)) == NULL ||
            (argv[2] = cbor_encode_pin_auth(dev, ecdh, enc)) == NULL ||
-           (argv[3] = cbor_encode_pin_opt(dev)) == NULL) {
+           (prot != 1 && (argv[3] = cbor_build_uint8(prot)) == NULL)) {
                fido_log_debug("%s: cbor encode", __func__);
                r = FIDO_ERR_INTERNAL;
                goto fail;

```

Does that make sense?

-p.

niryanowsky:
Make sense, and working, cool.
It will be out as part of 1.9.0? when it should be out ?

martelletto:
Thank you for testing. Yes, the diff above will be part of 1.9.0, ETA November 2021.

PR #437, opened 2021-10-02
iso7816: Avoid storing pointers in a packed structure

jrtc27:
On CHERI, and thus Arm's experimental Morello prototype architecture,
pointers are represented as capabilities, which are unforgeable bounded
pointers, providing always-on fine-grained spatial memory safety. The
unforgeability is enforced through the use of tagged memory, with one
validity tag bit per capability-sized-and-aligned word in memory. This
means that storing a pointer to an unaligned location, which is not
guaranteed to work per the C standard, either traps or results in the
capability losing its tag (and thus never being dereferenceable again),
depending on how exactly the store is done (specifically, whether a
capability store or memcpy is used).

However, iso7816 itself does not need to be packed, and doing so likely
causes inefficiencies on existing architectures. The iso7816_header_t
member is packed, and the flexible payload array is a uint8_t (which by
definition has no padding bits and is exactly 8 bits in size and, since
CHAR_BITS must be at least 8, its existence implies that it has the same
representation as unsigned char, and that it has size and alignment 1)
so there will never be any padding inserted between header and payload.

### DIFF
--- a/src/iso7816.h
+++ b/src/iso7816.h
@@ -27,14 +27,13 @@ struct iso7816_header {
 	uint8_t lc3;
 })
 
-PACKED_TYPE(iso7816_apdu_t,
-struct iso7816_apdu {
+typedef struct iso7816_apdu {
 	size_t            alloc_len;
 	uint16_t          payload_len;
 	uint8_t          *payload_ptr;
 	iso7816_header_t  header;
 	uint8_t           payload[];
-})
+} iso7816_apdu_t;
 
 const unsigned char *iso7816_ptr(const iso7816_apdu_t *);
 int iso7816_add(iso7816_apdu_t *, const void *, size_t);
--- a/src/iso7816.c
+++ b/src/iso7816.c
@@ -59,6 +59,6 @@ iso7816_ptr(const iso7816_apdu_t *apdu)
 size_t
 iso7816_len(const iso7816_apdu_t *apdu)
 {
-	return apdu->alloc_len - sizeof(apdu->alloc_len) -
-	    sizeof(apdu->payload_len) - sizeof(apdu->payload_ptr);
+	return apdu->alloc_len - offsetof(iso7816_apdu_t, header) -
+	    (sizeof(iso7816_apdu_t) - offsetof(iso7816_apdu_t, payload));
 }


src/iso7816.h
```diff
@@ -27,14 +27,13 @@ struct iso7816_header {
 	uint8_t lc3;
 })
```
jrtc27, line 28: Note that, by similar logic, this also doesn't actually need to be packed, but doing so is a harmless no-op and reduces churn

src/iso7816.h
```diff
@@ -27,14 +27,13 @@ struct iso7816_header {
...
-PACKED_TYPE(iso7816_apdu_t,
-struct iso7816_apdu {
+typedef struct iso7816_apdu {
 	size_t            alloc_len;
 	uint16_t          payload_len;
 	uint8_t          *payload_ptr;
 	iso7816_header_t  header;
 	uint8_t           payload[];
-})
+} iso7816_apdu_t;
```
martelletto, line 36: On platforms where `size_t` is 4 or 8 bytes, this causes one byte of padding to be added to the end of `iso7816_apdu_t`, which breaks the calculation in `iso7816_len()`. I suggest accounting for this extra byte when we set `alloc_len` in `iso7816_new()`.

martelletto (changes requested):

src/iso7816.h
jrtc27: Ah indeed, and it's even greater on CHERI as our pointers on 64-bit architectures are 128-bit, so we end up with 9 padding byte.

martelletto: Looking at the resulting diff, it strikes me as simpler to (potentially) overallocate in `iso7816_new()` and have the adjustment logic entirely in `iso7816_len()`. I've just pushed a commit to that effect. Could I kindly ask you to squash it with yours? Provided you agree it's a sensible approach, of course.

jrtc27: Yeah that's fine by me, overallocates a bit but is simpler to reason about. Also means `iso7816_len` is now always precise again, whereas my version would report it was longer than needed for small messages; don't know if that would have been a problem.

martelletto (approved):

martelletto:
LGTM. Thank you!

LDVG (approved):

kongeo (approved):

PR #444, opened 2021-10-17
cmake: Add -Werror when checking host compiler for -fstack-protector

glaubitz:
Since calling GCC with -fstack-protector will just generate a warning
when it doesn't support the flag on a given host architecture, we need
to add -Werror so the warning turns into an error and cause the cmake
check HAVE_STACK_PROTECTOR_ALL to fail when the flag is not supported.

fixes gh#443

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -80,7 +80,7 @@ if(NOT MSVC)
 endif()
 
 check_c_compiler_flag("-Wshorten-64-to-32" HAVE_SHORTEN_64_TO_32)
-check_c_compiler_flag("-fstack-protector-all" HAVE_STACK_PROTECTOR_ALL)
+check_c_compiler_flag("-Werror -fstack-protector-all" HAVE_STACK_PROTECTOR_ALL)
 
 check_include_files(cbor.h HAVE_CBOR_H)
 check_include_files(endian.h HAVE_ENDIAN_H)


jrtc27:
The trouble with this kind of a approach in general is that configure-time checks often aren't warning clean, and so -Werror can turn up false negatives (in that it believes certain features don't work, but that's only because the check was slightly dodgy and triggered a warning: unused variable, prototype-less function, command-line flag that only applies when linking, etc., I've seen all those). In this case I don't think there's a way to avoid that potential problem, though; GCC doesn't even seem to apply a warning category to that warning so you can't just do some kind of -Werror=stack-protector-not-supported thing (whereas for CHERI Clang we at least categorise it under -Woption-ignored).

martelletto:
Looks reasonable to me. To @jrtc27's point, there is a chance of false negatives, but that's acceptable in this context -- the intention is to opportunistically enable `-fstack-protector-all`. For reference, this is the [check](https://github.com/Kitware/CMake/blob/c19b10883/Modules/Internal/CheckCompilerFlag.cmake#L15-L17), and this is the [list of regular expressions](https://github.com/Kitware/CMake/blob/c19b10883/Modules/CMakeCheckCompilerFlagCommonPatterns.cmake#L11-L34) used by CMake. If anyone is feeling particularly charitable, there might be a CMake PR in store for you.

PR #475, opened 2021-11-29
winhello: fix function typedefs for 32 bit x86

github-cygwin:
Missing WINAPI leads to broken stacks after calling webauthn functions
on 32 bit Cygwin.  The reason is that GCC defaults to the cdecl
calling convention on 32 bit x86, while WINAPI is equivalent to the
stdcall calling convention.

### DIFF
--- a/src/winhello.c
+++ b/src/winhello.c
@@ -50,21 +50,21 @@ struct winhello_cred {
 	wchar_t						*display_name;
 };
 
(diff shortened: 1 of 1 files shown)


martelletto (approved):

LDVG (approved):

kongeo (approved):

martelletto:
Thank you!

Issue #476, opened 2021-12-04
Build is broken on NetBSD without _NETBSD_SOURCE


riastradh:
**What version of libfido2 are you using?**
master (post-1.9.0)

**What operating system are you running?**
NetBSD

**What application are you using in conjunction with libfido2?**
N/A

**How does the problem manifest itself?**
build fails

```
cd /home/riastradh/crypto/webauthn/libfido2/build/src && /usr/bin/cc -DHAVE_ARC4RANDOM_BUF -DHAVE_CLOCK_GETTIME -DHAVE_DEV_URANDOM -DHAVE_ENDIAN_H -DHAVE_ERR_H -DHAVE_GETLINE -DHAVE_GETOPT -DHAVE_GETPAGESIZE -DHAVE_OPENSSLV_H -DHAVE_SIGNAL_H -DHAVE_STRLCAT -DHAVE_STRLCPY -DHAVE_STRSEP -DHAVE_SYSCONF -DHAVE_TIMESPECSUB -DHAVE_UNISTD_H -DTLS=__thread -D_FIDO_INTERNAL -D_FIDO_MAJOR=1 -D_FIDO_MINOR=10 -D_FIDO_PATCH=0 -I/home/riastradh/crypto/webauthn/libfido2/src -I/pkg/2021Q3/include -D_POSIX_C_SOURCE=200809L -D_BSD_SOURCE -std=c99  -fPIC -Wall -Wextra -Werror -Wshadow -Wcast-qual -Wwrite-strings -Wmissing-prototypes -Wbad-function-cast -pedantic -pedantic-errors -fstack-protector-all -Wno-unused-result -Wconversion -Wsign-conversion -MD -MT src/CMakeFiles/fido2.dir/aes256.c.o -MF CMakeFiles/fido2.dir/aes256.c.o.d -o CMakeFiles/fido2.dir/aes256.c.o -c /home/riastradh/crypto/webauthn/libfido2/src/aes256.c
In file included from /home/riastradh/crypto/webauthn/libfido2/src/fido.h:24:0,
                 from /home/riastradh/crypto/webauthn/libfido2/src/aes256.c:7:
/home/riastradh/crypto/webauthn/libfido2/src/blob.h:31:40: error: unknown type name 'u_char'
 int fido_blob_set(fido_blob_t *, const u_char *, size_t);
                                        ^~~~~~
```

**Is the problem reproducible?**
100%

**What are the steps that lead to the problem?**
```
mkdir build
cd build
[...]

PR #477, opened 2021-12-04
Define _NETBSD_SOURCE when building on NetBSD.

riastradh:
Otherwise, various nonstandard but common things that libfido2
assumes are defined, like u_char and u_int in <sys/types.h>, are
unavailable.

fix https://github.com/Yubico/libfido2/issues/476


martelletto:
Thank you and apologies for the breakage.

Issue #465, opened 2021-11-14
Support custom transport functions


riastradh:
I would like to automatically test an application that uses libfido2 to make and verify credentials.  To this end, I would like to drop in a software U2F/FIDO emulator that can be used with as much of the unmodified libfido2 API as possible given a simple opt-in by an application.

**Describe the feature you would like.**
1. A way to pass a handle through fido transport functions.

   Currently fido io functions (fido_dev_io_t) have an opaque handle they can use internally, but not fido transport functions (fido_dev_handle_t).  The one use of fido transport functions inside libfido2 (nfc_linux.c) uses the io handle, but since the structure of fido_dev_t is opaque to applications outside libfido2, so custom fido transport functions can't get at the io handle either.  So custom fido transport functions are not very useful outside libfido2 (they would have to rely on static storage for what amounts to a single global device handle in the program).  Applications can use custom fido io functions instead, but that adds some extra encapsulation and decapsulation hassle that libfido2 could obviate with custom fido transport functions instead.

   Because custom fido transport functions are not very useful as is, I suggest simply changing the signatures of the transport tx and rx functions to pass an opaque handle as an argument -- either the io handle, or another handle that can be set with a new API function or parameter:

   ```
   typedef int   fido_dev_rx_t(struct fido_dev *, void *handle, uint8_t, unsigned char *, size_t, int);
   typedef int   fido_dev_tx_t(struct fido_dev *, void *handle, uint8_t, const unsigned char *, size_t);
   ```

   While here, it would be nice if fido_dev_set_sigmask worked with custom fido io or transport functions.

2. Export `fido_dev_new_with_info` and `fido_dev_open_with_info`.

   Currently these are defined, but not exported in src/export.gnu &c., so there is no way for users of the library to request the io or transport functions in a fido_dev_into_t be used in a fido_dev_t, and to use the path in a fido_dev_info_t, you have to write `fido_dev_open(dev, fido_dev_info_path(devinfo))`.

2. A way to fill fido_dev_info_t with custom path/vendor/product/transport/io.

   My application uses fido_dev_info_manifest to list the available devices, and then iterates over them (concurrently in threads) to use the devices with fido_dev_info_path, fido_dev_new, fido_dev_open, &c.  With (2) above, it would use fido_dev_new_with_info and fido_dev_open_with_info instead, and automatically get any custom io or transport functions.

[...]

PR #478, opened 2021-12-04
Add missing support for custom transport functions

riastradh:
fix https://github.com/Yubico/libfido2/issues/465

For the moment, I didn't provide any way to use _only_ custom transport functions -- it is still necessary to combine custom I/O functions (open and close, but not read or write) with custom transport functions (tx and rx).  This serves for my purposes, but it might be nice if the transport layer operated independently of the I/O layer.  Also, for the moment, the only way to set the device handle is via the I/O open function, which is only passed a path as an argument -- again, adequate for my purposes, but other applications might want simpler control from the caller over the device handle.  (Of course, you could snprintf("%p") and then sscanf it back!)

### DIFF
--- a/man/fido_dev_set_io_functions.3
+++ b/man/fido_dev_set_io_functions.3
@@ -8,7 +8,9 @@
 .Sh NAME
 .Nm fido_dev_set_io_functions ,
 .Nm fido_dev_set_sigmask ,
-.Nm fido_dev_set_timeout
+.Nm fido_dev_set_timeout ,
+.Nm fido_dev_set_transport_functions ,
+.Nm fido_dev_io_handle
 .Nd FIDO 2 device I/O interface
 .Sh SYNOPSIS
 .In fido.h
@@ -30,13 +32,28 @@ typedef int fido_sigset_t;
 #else
 typedef sigset_t fido_sigset_t;
 #endif
+
+typedef int   fido_dev_rx_t(struct fido_dev *,
+                  uint8_t, unsigned char *, size_t, int);
+typedef int   fido_dev_tx_t(struct fido_dev *,
+                  uint8_t, const unsigned char *, size_t);
+
+typedef struct fido_dev_transport {
+	fido_dev_rx_t *rx;
+	fido_dev_tx_t *tx;
+} fido_dev_transport_t;
 .Ed
+.Pp
 .Ft int
 .Fn fido_dev_set_io_functions "fido_dev_t *dev" "const fido_dev_io_t *io"
 .Ft int
 .Fn fido_dev_set_sigmask "fido_dev_t *dev" "const fido_sigset_t *sigmask"
 .Ft int
 .Fn fido_dev_set_timeout "fido_dev_t *dev" "int ms"
+.Ft int
+.Fn fido_dev_set_transport_functions "fido_dev_t *dev" "const fido_dev_transport_t *t"
+.Ft void *
+.Fn fido_dev_io_handle "fido_dev_t *dev"
 .Sh DESCRIPTION
 The
 .Fn fido_dev_set_io_functions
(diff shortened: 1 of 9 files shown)


src/hid.c
```diff
@@ -148,6 +148,36 @@ fido_dev_info_ptr(const fido_dev_info_t *devlist, size_t i)
 	return (&devlist[i]);
 }
 
+int
+fido_dev_info_set(fido_dev_info_t *devlist, size_t i,
```
LDVG, line 152: Suggest renaming to `fido_dev_set_info()` for consistency with other setters in `libfido2`.

src/hid.c
```diff
@@ -148,6 +148,36 @@ fido_dev_info_ptr(const fido_dev_info_t *devlist, size_t i)
...
+    const char *path, const char *manufacturer, const char *product,
+    const fido_dev_io_t *io, const fido_dev_transport_t *transport)
+{
+	char *path_copy = NULL, *manu_copy = NULL, *prod_copy = NULL;
+	int r = FIDO_ERR_INTERNAL;
+
+	if ((path && (path_copy = strdup(path)) == NULL) ||
```
LDVG, line 159: `fido_dev_open_with_info()` does not accept a `NULL` path, perhaps we should return `FIDO_ERR_INVALID_ARGUMENT` for that case here?

For devices without `manufacturer` and `product` strings, we use an empty string rather than `NULL` (see #414). Perhaps it'd make sense to align with that behavior here as well.

src/hid.c
```diff
@@ -148,6 +148,36 @@ fido_dev_info_ptr(const fido_dev_info_t *devlist, size_t i)
...
+	devlist[i].path = path_copy;
+	devlist[i].manufacturer = manu_copy;
+	devlist[i].product = prod_copy;
+	if (io)
+		devlist[i].io = *io;
+	if (transport)
+		devlist[i].transport = *transport;
+	r = FIDO_OK;
+
+out:	if (r) {
```
LDVG, line 173: Prefer having the label on its own line and checking the return value against `FIDO_OK` explicitly.

src/hid.c
riastradh: The name `fido_dev_set_info` suggests to me that it operates on a `fido_dev_t`, but that's not what this does -- it operates on a `fido_dev_info_t` array, like `fido_dev_info_new` returns.  Maybe `fido_dev_info_set_parameters` or `fido_dev_info_set_entry` or `fido_dev_info_init`, if `fido_dev_info_set` still seems wrong?

LDVG: You are absolutely correct; the current name is fine by me. :+1:

man/fido_dev_set_io_functions.3
```diff
@@ -148,9 +165,55 @@ This is the default behaviour.
...
+and
+.Dv write
+functions of the I/O handlers.
+However, the I/O handlers must still be specified to open and close the
+device.
+The
+.Fn fido_dev_io_handle
+function returns the opaque pointer returned by the
+.Dv open
+function of the I/O handlers, so it carry information to the transport
```
LDVG, line 211: Do we need the trailing part about transport handlers to document this function?

riastradh: I split it into a separate sentence.  I thought it would be worthwhile to explain why the function exists in connection to fido_dev_set_transport_functions, since it's not really necessary for fido_dev_set_io_functions on its own.  Do you think the motivation is not worth keeping?

LDVG: Your updated version with the expanded rationale looks good to me, thanks!

LDVG (approved):
This looks good to me.

WDYT @martelletto; @kongeo?

riastradh:
FYI, rebased and squashed down to three independently functional commits.  No change in the result.

martelletto (approved):

kongeo (approved):

PR #502, opened 2022-01-06
Adding static lib symbols to package folders

anmenaga:
This change adds following symbol files for static libraries in the output package folders:
`crypto-46.pdb, cbor.pdb, zlib.pdb, fido2.pdb`


martelletto:
Thank you!

PR #523, opened 2022-02-14
Advertise "uv" capability, not "clientPin".

github-cygwin:
Per the CTAP protocol the "uv" capability indicates that the device is
capable of verifying the user within itself, while "clientPin" indicates
that the device is capable of accepting a PIN from the client.  WinHello
only supports "uv", but not "clientPin", but current libfido2 advertises
"clientPin", not "uv".  Change this in fido_winhello_get_cbor_info.

Signed-off-by: Corinna Vinschen <vinschen@redhat.com>


LDVG (approved):

martelletto:
Thank you!

PR #524, opened 2022-02-17
dev: restrict visibility of fido_dev_set_sigmask

LDVG:
On UNIX-like systems, an application compiled using strict ISO C may
fail to build if it includes libfido2 headers due to `sigset_t` being
a POSIX extension. Determine whether `sigset_t` is defined by checking
for a known POSIX macro from the same header, hide the declaration of
`fido_dev_set_sigmask()` and `fido_sigset_t` otherwise.


martelletto (approved):
LGTM. Thank you!

PR #531, opened 2022-02-27
udev: add SoloKey v2 ID

bjacquin:
See: https://github.com/solokeys/solo2
See: https://github.com/solokeys/solo2/blob/main/70-solo-bee.rules

Cc @conorpp


martelletto:
Thank you!

PR #543, opened 2022-03-15
cmake: drop FUZZ_PCSC

martelletto:
the FUZZ_PCSC cmake option was a stopgap solution to keep oss-fuzz builds running without pcsclite headers. with the headers in place, FUZZ_PCSC is no longer needed.


LDVG (approved):

martelletto:
Further changes are needed in oss-fuzz. I'm marking this PR as a draft.

Issue #546, opened 2022-03-21
es256_pk_from_EC_KEY doesn't support coordinates using less than 32 bytes


agalluzzi-cybele:
When key is ES256, public key cbor map stores coordinates as 32 bytes values. In order to create an _es256_pk_t_ I was loading an EC_KEY storing parameters readed from public key cbor map in BN values which will be later loaded in EC_KEY using OpenSSL methods:

`BN_bin2bn(cbor_bytestring_handle(pkMap[i].value), cbor_bytestring_length(pkMap[i].value), x);`

I had an issue when one of the coordinates was less than 32 bytes long, for example, because there was a leading zero. In my case, `cbor_bytestring_handle(pkMap[i].value)` for key -2 was:

    [0]: 0 '\0'
    [1]: 51 '3'
    [2]: 196 'Ä'
    [3]: 103 'g'
    [4]: 59 ';'
    [5]: 13 '\r'
    [6]: 77 'M'
    [7]: 116 't'
    [8]: 255 'ÿ'
    [9]: 207 'Ï'
    [10]: 208 'Ð'
    [11]: 143 ''
    [12]: 131 'ƒ'
    [13]: 229 'å'
    [14]: 58 ':'
    [15]: 129 ''
    [16]: 26 '\x1a'
    [17]: 171 '«'
    [18]: 91 '['
    [19]: 90 'Z'
    [20]: 226 'â'
    [21]: 63 '?'
    [22]: 59 ';'
    [23]: 117 'u'
    [24]: 109 'm'
    [25]: 51 '3'
    [26]: 60 '<'
    [27]: 173 '­'
    [28]: 21 '\x15'
    [29]: 178 '²'
    [30]: 89 'Y'
    [31]: 29 '\x1d'

After loading the value in the EC_KEY, I was calling

```
es256_pk_t* npk = es256_pk_new();
es256_pk_from_EC_KEY(npk, ec);
```

When inspecting `npk->x` there was no leading zero, and that changed all the curve and validation failed. In order to fix that I had to change _es256_pk_from_EC_KEY_ to offset _BN_bn2bin_ to the right location:

```
int
es256_pk_from_EC_KEY(es256_pk_t *pk, const EC_KEY *ec)
{
	BN_CTX		*bnctx = NULL;
	BIGNUM		*x = NULL;
	BIGNUM		*y = NULL;
	const EC_POINT	*q = NULL;
	const EC_GROUP	*g = NULL;
[...]

PR #547, opened 2022-03-22
Fix es256_pk_from_EC_KEY()  on public keys with short coordinates

martelletto:
Reported in #546. Seems to happen about 0.5% of the time (assuming random keys).


martelletto:
(I am placing the regression test before the fix to demonstrate the problem.)

LDVG (approved):

kongeo (approved):

PR #560, opened 2022-04-12
compress: use raw deflate/rfc1951

LDVG:
Backwards compatibility kept by attempting zlib/rfc1950 decompression
first. Joint work with @martelletto.


regress/compress.c
```diff
@@ -109,42 +109,158 @@ static /* const */ unsigned char zlib_blob[694] = {
...
+	assert(fido_compress(&out, &in) == FIDO_OK);
+
+	in.ptr = out.ptr;
+	in.len = out.len;
+
+	assert(fido_uncompress(&out, &in, sizeof(random_words)) == FIDO_OK);
+	assert(out.len == sizeof(random_words));
+	assert(memcmp(out.ptr, random_words, out.len) == 0);
+
+	free(out.ptr);
```
martelletto, line 253: should we free(in.ptr) here?

martelletto (approved):
left a small comment regarding a regression test; lgtm otherwise

regress/compress.c
LDVG: yes, fixed. thanks!

kongeo (approved):

Issue #247, opened 2020-12-08
Android support


jpalus:
Is there any chance on adding Android to supported platforms? It would be really cool to use it with openssh inside Termux in a similar fashion to TermBot. Not sure if Android + NFC is in scope of this library though.

martelletto:
NFC on Linux is being worked on, and may land before the end of the year. Whether it will be enough to support Android is unclear. Android as a supported platform is unlikely to happen unless there is code contribution (and maintenance) by a third party.

cyberbob:
I'm using ACR1252U and Yubikey 5 NFC on OSX, is this the same code base as for Linux ?

martelletto:
Unfortunately, no. I can try to have a look at NFC on macOS at some point next year, but can't promise anything. Sorry.

martelletto:
Experimental support for NFC on Linux has been merged in https://github.com/Yubico/libfido2/commit/0a1a5ce.

awakecoding:
Any update on this? I noticed that the Linux port requires libudev, which is not available on Android AFAIK. If libudev could be fixed, what else would be required to get libfido2 up and running on Android? Are there devices that should normally work with an Android phone?

martelletto:
We are not aware of efforts in this area. I don't know enough about Android to say what would be required, but I expect libudev not to be a problem, even if it isn't available. As to devices, any USB-C or NFC device should work, at least in theory.

awakecoding:
@martelletto I'll cleanup my patches another day, but I just wanted to report that I've managed to get it build for Android using [libudev-zero](https://github.com/illiliti/libudev-zero). explicit_bzero / bzero is missing on Android, but explicit_bzero is defined in libcrypto, so I've tweaked the code a little bit. Other than that, it's really like a Linux build. My CMake build system patches are here: https://github.com/awakecoding/libfido2/tree/devolutions

martelletto:
@awakecoding That's great news! Don't worry about cleaning up too much; anything that works is a good start, and we will happily take it from there. Did you get to test the resulting build? I was under the impression that Android applications would not be able to access /dev/hidraw* due to sandboxing.

awakecoding:
[...]

PR #571, opened 2022-04-25
Android build fixes

awakecoding:
Fix various issues to get [libfido2 to build on Android](https://github.com/Yubico/libfido2/issues/247). I have no idea if it works, but getting it to build is the first step.

- Add BUILD_TESTS option to facilitate disabling the tests on platforms where it causes problem
- Add CRYPTO_EXPLICIT_BZERO to use explicit_bzero symbol from libcrypto (Android libc doesn't have it)
- Detect libudev-zero as a replacement for libudev on Android (could be used for Linux in the future)
- Fix using proper zlib cmake target to link against zlib (required for non-default install locations)

I prebuilt and inject all dependencies as static libraries in my build, and I have a few more patches on my side to make it easier to inject non-default libraries. However, this would require larger changes that should probably come in a second pull request.


awakecoding:
@martelletto can you re-run the last build? I think the Windows build failure is just a fluke "The hosted runner: GitHub Actions 13 lost communication with the server. Anything in your workflow that terminates the runner process, starves it for CPU/Memory, or blocks its network access can cause this error."

martelletto:
Thank you for the PR! To merge it, we need to:

- [ ] set up a CI runner for Android to ensure there are no regressions;
- [ ] test on Android to find out what else we need.

I unfortunately won't be able to look into it in the near future, but am happy to provide feedback if anyone is interested.

martelletto:
Thank you for the PR! To merge it, we need to:

- [ ] set up a CI runner for Android to ensure there are no regressions;
- [ ] test on Android to find out what else we need.

I unfortunately won't be able to look into it in the near future, but am happy to provide feedback if anyone is interested.

awakecoding:
@martelletto isn't it possible to simply merge this first batch of Android build fixes, considering it is truly meant to make it build, not really make it work yet? In my case I only wish to build libfido2 for Android in order to build Win32-OpenSSH for Android, simply because it depends on it. I don't even know if it's functional, but at least I can build Win32-OpenSSH now, and it should be functional even without FIDO2 support. I haven't tried building for iOS yet, but I was thinking of doing the same, at least for libssh because iOS doesn't let you build and run executables easily.

martelletto:
@awakecoding, unless it is an absolute blocker for you, our preference would be to wait and test before merging.

awakecoding:
@martelletto it is not an **absolute** blocker for me, but this is as far as my current contribution will go. If you don't merge it, it will rot on a branch for years until somebody else comes along to decide and do the work all the way and helps you set up automated testing for Android at the same time. My downstream patches are currently based off the Microsoft libfido2 fork which already modifies the build system, I am trying to reduce my reliance on custom patches such that I could easily fork from the original upstream repository (this one) in the future.

The current pull request is not meant to close the ["Android support" feature request](https://github.com/Yubico/libfido2/issues/247), as it only makes it build for now. I don't have the knowledge or experience to push it any further than that, but someone else eventually will. When this happens, we don't want duplicated efforts to just get it to build initially. The primary value for me, however, is that even if not officially supported, I can correctly build Win32-OpenSSH for Android, despite not truly supporting FIDO2 at this point.

martelletto:
@awakecoding, I understand your point, and truly hope you understand ours too. Given it's not the functionality you are after, would you agree this is something that would ideally be tackled in Win32-OpenSSH? Upstream OpenSSH can be built without libfido2 (--disable-security-key). I suppose something similar could be implemented in Win32-OpenSSH. Happy to give it a shot next week if you think it is a good idea.

PR #579, opened 2022-05-06
regress: link/run against chosen library

silkeh:
- Compile `regress` against `${_FIDO2_LIBRARY}` so it links against the shared library when building with `BUILD_STATIC_LIBS=OFF`.
- Export the symbols required for `regress/compress`.
- Run the Linux CI tests for both static and dynamic libraries.

### DIFF
--- a/regress/CMakeLists.txt
+++ b/regress/CMakeLists.txt
@@ -4,11 +4,11 @@
 
 add_custom_target(regress)
 
-macro(add_regress_test NAME SOURCES)
+macro(add_regress_test NAME SOURCES LIB)
 	add_executable(${NAME} ${SOURCES})
-	target_link_libraries(${NAME} fido2)
 	add_test(${NAME} ${NAME})
 	add_dependencies(regress ${NAME})
+	target_link_libraries(${NAME} ${LIB})
 endmacro()
 
 if(MSVC AND BUILD_SHARED_LIBS)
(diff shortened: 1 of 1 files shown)


martelletto:
Thank you for the PR. The `fido_compress`, `fido_uncompress`, and `fido_sha256` functions are not part of the official API, and therefore should not be exported -- doing so defeats the purpose of export lists. If that's where we want to go, then we should probably drop export lists altogether, which I don't think is a good idea.

An alternative solution is to define a separate export list for the regression tests, like we do for the libFuzzer harnesses (https://github.com/Yubico/libfido2/blob/main/fuzz/export.gnu). That would require linking a separate shared object and pointing the regression tests at it, at which point it strikes me as simpler to just build a static version of the library.

silkeh:
> Thank you for the PR. The `fido_compress`, `fido_uncompress`, and `fido_sha256` functions are not part of the official API, and therefore should not be exported -- doing so defeats the purpose of export lists. If that's where we want to go, then we should probably drop export lists altogether, which I don't think is a good idea.

Thanks for the clarification! In that case it seems a bit strange to me that the regression tests use unexported functionality.
Shouldn't the regression tests only use exported functionality?

> An alternative solution is to define a separate export list for the regression tests, like we do for the libFuzzer harnesses (https://github.com/Yubico/libfido2/blob/main/fuzz/export.gnu). That would require linking a separate shared object and pointing the regression tests at it, at which point it strikes me as simpler to just build a static version of the library.

In both cases the tests would run against a different build (either static or dynamic) than the one that is actually requested and installed by `make install`, which makes it possible (though unlikely) to have issues in the installed library but not in the test.

Perhaps as a simple compromise we could disable affected tests when built against the shared library?

martelletto:
Hi @silkeh,

Apologies for the delay. I have been a bit busy.

> Thanks for the clarification! In that case it seems a bit strange to me that the regression tests use unexported functionality. Shouldn't the regression tests only use exported functionality?

This is functionality that would otherwise be hard to exercise without an actual device and everything it entails (touch, PIN, etc).

> Perhaps as a simple compromise we could disable affected tests when built against the shared library?

Sounds good to me. We would need to skip regress/compress.c.

-p.

silkeh:
> Apologies for the delay. I have been a bit busy.

No worries, I happened to be unavailable for a bit as well!

> > Perhaps as a simple compromise we could disable affected tests when built against the shared library?
> 
> Sounds good to me. We would need to skip regress/compress.c.

Ok! I have updated the PR with a skip of `compress.c` when only the shared libraries are built.

There are some remaining CI failures for Cygwin. My guess is that the library cannot be found when executing the tests, but I don't have a Windows installation to play around with to actually debug this properly. Suggestions welcome!

martelletto:
@silkeh, apologies for the delay. I have pushed a commit to address Cygwin and to (hopefully) simplify regress/CMakeLists.txt. When you have a chance, please let me know what you think. Thank you.

silkeh:
Looks good, thanks!

martelletto:
Thank you. I've dropped the CI bits for now (further work to happen in a separate, future PR) and reworded the commit message a little.

PR #604, opened 2022-07-07
hid_linux: reopen hidraw(4) after flock(2)

martelletto:
failure to do so will cause a process to observe hid frames exchanged by a previous holder of the lock. while there's code in rx_preamble() to skip frames addressed to a different cid (allowing for a degree of synchronisation), it is unclear what should happen to ctap broadcast frames not addressed to the current process. debugged and discussed with @LDVG.

### DIFF
--- a/src/hid_linux.c
+++ b/src/hid_linux.c
@@ -244,6 +244,10 @@ fido_hid_open(const char *path)
 	struct hidraw_report_descriptor *hrd;
 	struct timespec tv_pause;
 	long interval_ms, retries = 0;
+	bool looped;
+
+retry:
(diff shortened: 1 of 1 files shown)


martelletto:
prompted by and possibly related to https://github.com/systemd/systemd/issues/23889

LDVG (approved):

kongeo (approved):

Thaodan:
Will you backport this to the latest release version?

martelletto:
This fix will be part of libfido2 1.12.0, to be released at some point in September.

PR #645, opened 2022-11-11
Disable EdDSA compat for LibreSSL 3.7

botovq:
LibreSSL 3.7 will ship with Ed25519 support in EVP so the compatibility functions result in a symbol collision. Disable them.


LDVG (approved):

LDVG:
Thank you!

PR #662, opened 2023-01-07
support different fuzzing engines than libfuzzer

martelletto:
this PR allows libfido2's fuzz harnesses to be built with anything providing a libfuzzer-like API. these include afl, honggfuzz, and centipede, as packaged by oss-fuzz.


LDVG (approved):
LGTM! Tested locally using the oss-fuzz helper scripts; builds and runs with libfuzzer, afl, and honggfuzz whereas centipede seem to have unrelated linking issues.

kongeo (approved):

PR #665, opened 2023-01-11
Add support for unicode in winhello.c

ntwerdochlib:
Wrap parameter in the LoadLibrary() call with the TEXT macro so that the code can be used in projects supporting wide strings


kongeo (approved):

kongeo:
thank you!

PR #666, opened 2023-01-11
Replace the deprecated cmake subdirs call with add_subdirectory

ntwerdochlib:
The cmake subdirs() call was marked as deprecated in CMake 3.0.  This updates the  CMakeLists.txt file to use add_subdirectory()


LDVG:
Would you amending your commit to also replace the `subdirs(udev)` call on line 466?  Thanks!

ntwerdochlib:
Selective commits are always dangerous..  Will update the PR

LDVG:
Cheers!

PR #672, opened 2023-01-25
Removes always false integer comparison

James-ZHANG:
According to https://learn.microsoft.com/en-us/cpp/c-runtime-library/data-type-constants?view=msvc-170, `SIZE_MAX` is either `2^32-1` or `2^64-1`, and according to https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-dtyp/262627d8-3418-4627-9218-4ffe110850b2, `DWORD` is a 32-bit unsigned int. This implies a `DWORD > SIZE_MAX` comparison would always be false.

I make this finding in a Cygwin environment.

### DIFF
--- a/src/winhello.c
+++ b/src/winhello.c
@@ -472,10 +472,6 @@ unpack_assert_authdata(fido_assert_t *assert, const WEBAUTHN_ASSERTION *wa)
 {
 	int r;
 
-	if (wa->cbAuthenticatorData > SIZE_MAX) {
-		fido_log_debug("%s: cbAuthenticatorData", __func__);
-		return -1;
-	}
 	if ((r = fido_assert_set_authdata_raw(assert, 0, wa->pbAuthenticatorData,
 	    (size_t)wa->cbAuthenticatorData)) != FIDO_OK) {
 		fido_log_debug("%s: fido_assert_set_authdata_raw: %s", __func__,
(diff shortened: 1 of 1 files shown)


LDVG:
Thank you for the PR. If possible, we'd like to keep these checks in the spirit of always checking before casting. We also tried (explicitly) enabling `-Wtautological-compare` in our Cygwin pipeline (as far as I'm aware, `-Wall` should include it), [generating no warnings](https://github.com/Yubico/libfido2/actions/runs/4014068729/jobs/6894146592#step:3:680). What environment/compiler are you seeing this with?

James-ZHANG:
The flags we explicitly added is: `-Wall -Wextra -Werror -Wshadow -Wcast-qual
    -Wwrite-strings -Wmissing-prototypes
    -Wno-unused-result -Wno-unused-parameter -Wno-int-to-pointer-cast
    -Wno-pointer-to-int-cast
    -Wbad-function-cast -pedantic
    -pedantic-errors -fstack-protector-all`.

Our `gcc-core` cygwin package is of version `6.4.0-4`.

LDVG:
Thank you for the clarification. With the proposed changes, do you agree it would then make sense to also drop the corresponding explicit `(size_t)` casts so that `-Wconversion` would kick in if necessary?

James-ZHANG:
I agree, but I would appreciate it if you can first post a commit for us to double check it doesn't trigger additional compile warnings (it shouldn't, but just want to be safe...)

LDVG (approved):

kongeo (approved):

James-ZHANG:
We checked that the new commit works for our set of compiler flags.
We should be able to merge this, thanks for your cooperation.

LDVG:
> We checked that the new commit works for our set of compiler flags. We should be able to merge this, thanks for your cooperation.

Thank you for the contribution! Apologies for the slight delay.

PR #673, opened 2023-01-25
Avoids openssl name clashes on Windows

James-ZHANG:
See https://github.com/openssl/openssl/issues/1093 for the motivation.

### DIFF
--- a/src/winhello.c
+++ b/src/winhello.c
@@ -9,6 +9,14 @@
 
 #include <stdlib.h>
 #include <windows.h>
+#if defined(_WIN32) || defined(__CYGWIN__)
+	// Name clashes between windows.h and openssl
+	#undef X509_NAME
+	#undef X509_CERT_PAIR
+	#undef X509_EXTENSIONS
+	#undef OCSP_REQUEST
+	#undef OCSP_RESPONSE
+#endif
 
 #include "fido.h"
 #include "webauthn.h"


LDVG:
Sorry, I'm not sure I follow the problem as outlined by the linked issue. I was under the impression that these `#undef` directives are included in OpenSSL's own header files (e.g. [`ossl_typ.h` in 1.1.1 stable](
https://github.com/openssl/openssl/blob/adc3cfe452daa0fe51b8059929b2dafdc42737e4/include/openssl/ossl_typ.h#L66-L73)). Do you have an example of an environment in which our build fails?

James-ZHANG:
I built libfido2 on Cygwin, and Cygwin's gcc does not define `_WIN32`: https://cygwin.com/pipermail/cygwin-developers/2012-July/010732.html.

Maybe the proper fix is to upstream the `__CYGWIN__` check to OpenSSL, then?

LDVG:
Thank you for the clarification. I'm still unsure exactly how this problem materializes in your build.

From my understanding, these symbols are defined in `wincrypt.h` which is included by `windows.h` [unless `WIN32_LEAN_AND_MEAN` is defined](https://learn.microsoft.com/en-us/windows/win32/winprog/using-the-windows-headers#faster-builds-with-smaller-header-files); libfido2 defines `WIN32_LEAN_AND_MEAN` for Cygwin builds so there should not be any name clashes:

https://github.com/Yubico/libfido2/blob/fe5aa1d2ee1e698345b56feef9325cd44c45d9c2/CMakeLists.txt#L60-L66

Am I missing something?

James-ZHANG:
Ah, that's exactly what caused my build to fail: we use a constrained build environment where we do not have cmake in our Cygwin and thus have to use a custom Makefile to build libfido2. Naturally, we missed this part in the upstream.

After adding `-DWIN32_LEAN_AND_MEAN -D_WIN32_WINNT=0x0600` in our Makefile my build works. 

Closing this PR. Thanks for your pointer :).

PR #700, opened 2023-06-03
Extend FIDO2 BLE support also for Linux

akemnade:
For Windows it was already added via gh#336,
so let's also add it for Linux.
Unpaired devices are ignored, the user has to pair independently of libfido use using the bluetooth manager provided by the desktop environment.


LDVG:
Thank you for your contribution! Before having a more in-depth look, may I ask what authenticator you are testing this against?

akemnade:
This was tested with:

eSecu FIDO2 pro (passkey printed on device, 20 byte control point size, fragmentation often needed, bluetooth turned off after a short time
of inactivity)
AirID2 Mini (full Numerical comparison for pairing, >200 byte control point size, so fragmentation rarely needed),

akemnade:
about the ci-failures:
cifuzz: it does not install dependencies correctly, and they are not specified in this repo, so I cannot do anything there easily.
The other fuzzer:
```
==16136==WARNING: MemorySanitizer: use-of-uninitialized-value
5096
    #0 0x7ff5d26bb286  (/lib/x86_64-linux-gnu/libsystemd.so.0+0x4b286) (BuildId: e45f7492c0f62251620378d7224ad0371a8d1f98)
5097
    #1 0x7ff5d26b2ae7 in sd_bus_start (/lib/x86_64-linux-gnu/libsystemd.so.0+0x42ae7) (BuildId: e45f7492c0f62251620378d7224ad0371a8d1f98)
5098
    #2 0x7ff5d26b2e88 in sd_bus_open_system_with_description (/lib/x86_64-linux-gnu/libsystemd.so.0+0x42e88) (BuildId: e45f7492c0f62251620378d7224ad0371a8d1f98)
5099
    #3 0x7ff5d26ba157 in sd_bus_default_system (/lib/x86_64-linux-gnu/libsystemd.so.0+0x4a157) (BuildId: e45f7492c0f62251620378d7224ad0371a8d1f98)
```
How to debug this? sd_bus_default_system just takes a pointer to a variable on the stack to write some output into this, so nothing scary. Can this be resolved to source code lines?

LDVG:
> How to debug this? sd_bus_default_system just takes a pointer to a variable on the stack to write some output into this, so nothing scary. Can this be resolved to source code lines?

MemorySanitizer requires the whole application to be instrumented -- libsystemd included.  

We'd likely need to mock a lot of these calls to sensibly fuzz your implementation.

src/bluetooth_linux.c
```diff
@@ -0,0 +1,673 @@
...
+	}
+
+	return FIDO_ERR_INTERNAL;
+}
+
+int
+fido_bluetooth_write(void *handle, const unsigned char *buf, size_t len)
+{
+	struct ble *dev = (struct ble *)handle;
+	sd_bus_message *send_msg;
```
LDVG, line 454: I believe a `sd_bus_message_unref` is needed here?

src/bluetooth_linux.c
```diff
@@ -0,0 +1,673 @@
...
+		reply = NULL;
+
+		if (0 > sd_bus_call_method(newdev->bus, "org.bluez", newdev->paths.service_revision,
+					DBUS_CHAR_IFACE, "ReadValue", NULL, &reply, "a{sv}", 0))
+			goto out;
+		if (0 > sd_bus_message_read(reply, "ay", 1, &revision))
+			goto out;
+
+		/* for simplicity, we allow now only FIDO2 */
+		if (!(revision & 0x20))
```
LDVG, line 358: The specification says

>  The client SHALL write a value to this characteristic with exactly 1 bit set before sending any FIDO commands unless u2fServiceRevision is present and U2F 1.0 compatibility is desired

which I believe we are missing?

src/bluetooth_linux.c
```diff
@@ -0,0 +1,673 @@
...
+
+#define DBUS_CHAR_IFACE "org.bluez.GattCharacteristic1"
+#define DBUS_DEV_IFACE "org.bluez.Device1"
+#define DBUS_SERVICE_IFACE "org.bluez.GattService1"
+#define DBUS_PROFILE_IFACE "org.bluez.GattProfile1"
+#define DBUS_ADAPTER_IFACE "org.bluez.Adapter1"
+#define DBUS_GATTMANAGER_IFACE "org.bluez.GattManager1"
+
+static sd_bus_slot *profile_slot = NULL;
+static sd_bus_slot *profile_manager_slot = NULL;
```
LDVG, line 22: Do we need these or can a floating bus slot be used instead?

src/bluetooth_linux.c
```diff
@@ -0,0 +1,673 @@
...
+{
+	const char *iface;
+	struct manifest_ctx *ctx = (struct manifest_ctx *) data;
+	if (sd_bus_message_read_basic(reply, 's', &iface) >= 0) {
+		if (!strcmp(iface, DBUS_ADAPTER_IFACE)) {
+			/* will also fail if adapter is not powered */
+			if (sd_bus_call_method(ctx->bus, "org.bluez", path, iface,
+					       "StartDiscovery", NULL, NULL, "") >= 0) {
+				ctx->scanning = true;
+			}
```
LDVG, line 184: Generally speaking, we've avoided waiting for devices in `fido_dev_info_manifest()`, do we need this or can we put the burden on the desktop environment to discover to the device?

src/bluetooth_linux.c
akemnade: yes, I will fix it

src/bluetooth_linux.c
akemnade: yes, it is missing

src/bluetooth_linux.c
akemnade: well, if libfido only looks for connected devices, than this stuff can be completely removed.

src/bluetooth_linux.c
akemnade: Well, I would prefer the connection being established when the fido usage started automatically (and disconnected if not used any longer), but I understand your concern.
bluetoothctl and the bluetooth ui from KDE allow manual connects, blueman not (but still allows pairing). I would agree to remove that connection/discovery method for now, it can be added later on

LDVG: > I would prefer the connection being established when the fido usage started automatically (and disconnected if not used any longer)

Without asking for any changes at this time -- would you agree that connecting and disconnecting should in that case preferably happen in `fido_dev_open()` and `fido_dev_close()` respectively? The application may not be calling `fido_dev_info_manifest()` at all, but instead rely on a previously known device path.

>  I would agree to remove that connection/discovery method for now, it can be added later on

Indeed, given the (very) asynchronous nature of BLE, we'll probably have to think a bit on what we'd like to do. Happy to return to it at some later point.

src/bluetooth_linux.c
```diff
@@ -0,0 +1,548 @@
...
+
+		if (0 > sd_bus_call_method(newdev->bus, "org.bluez", newdev->paths.service_revision,
+				DBUS_CHAR_IFACE, "WriteValue", NULL, NULL, "aya{sv}", 1, 0x20, 0))
+			goto out;
+
+		newdev->controlpoint_size = ((size_t)cp_len[0] << 8) + cp_len[1];
+		newdev->reply_buf = calloc(1, newdev->controlpoint_size);
+
+		if (0 > sd_bus_call_method(newdev->bus, "org.bluez", newdev->paths.status,
+					DBUS_CHAR_IFACE, "StartNotify", NULL, NULL, ""))
```
LDVG, line 299: Do you think using `AcquireNotify` instead would make sense?

We'd then be able to get rid of the `slot`, `reply_buf`, `reply_len` members of `struct ble` and be able to drop `read_cb` in its entirety. After setting up the file descriptor, we'd basically just need a call to  `fido_hid_unix_wait()`  followed by `read(2)` in `fido_bluetooth_read()`.

akemnade: yes, that is a really nice idea

src/bluetooth_linux.c
akemnade: well, I think all the device paths look a bit wonky for all kind of devices, but having that possibility to directly open a connection should be kept.
But not sure about all that. But maybe fido_dev_open should avoid doing anything consuming more time.  Probably the better place for anything consuming more time would be in fido_dev_info_manifest().

LDVG: > But maybe fido_dev_open should avoid doing anything consuming more time.

As an aside, there are situations in which `fido_dev_open()` may block, see for example the changes introduced by https://github.com/Yubico/libfido2/pull/350.

src/bluetooth_linux.c
```diff
@@ -0,0 +1,466 @@
...
+
+	ctx.devlist = devlist;
+	ctx.olen = olen;
+	ctx.ilen = ilen;
+	if (0>sd_bus_default_system(&bus))
+		return FIDO_ERR_INTERNAL;
+
+	ctx.bus = bus;
+	ret = sd_bus_call_method(bus, "org.bluez", "/", "org.freedesktop.DBus.ObjectManager",
+				 "GetManagedObjects", NULL, &reply, "");
```
LDVG, line 453: Nit: For second level indentation, we use four spaces.

```suggestion
	ret = sd_bus_call_method(bus, "org.bluez", "/", "org.freedesktop.DBus.ObjectManager",
	    "GetManagedObjects", NULL, &reply, "");
```

src/bluetooth_linux.c
```diff
@@ -0,0 +1,466 @@
...
+	reply = NULL;
+
+	if (newdev->paths.status &&
+	    newdev->paths.control_point &&
+	    newdev->paths.control_point_length &&
+	    newdev->paths.service_revision) {
+		uint8_t cp_len[2];
+		uint8_t revision;
+		if (0 > sd_bus_call_method(newdev->bus, "org.bluez", newdev->paths.control_point_length,
+					DBUS_CHAR_IFACE, "ReadValue", NULL, &reply, "a{sv}", 0))
```
LDVG, line 223: Nit: When breaking the line for a conditional, we typically align with the starting parenthesis.

```suggestion
		if (0 > sd_bus_call_method(newdev->bus, "org.bluez", newdev->paths.control_point_length,
		    DBUS_CHAR_IFACE, "ReadValue", NULL, &reply, "a{sv}", 0))
```

LDVG (commented):
This is starting to take shape!

I have left a set of comments which mostly aim to get this PR into a shape that is consistent with our existing code base. Note that some of the comments apply to more than one occurrence. I'd like to generally mention:

- We try to follow a basic KNF/style(9) format for our code.  While not critical, I left a couple of nitpicking comments on this.
- Some of the functions nest very deeply, we'd prefer if we could split some of the logic out to separate functions to alleviate this.
- We'd like for most return values to be checked and bubbled up to the calling function, and, where it makes sense, log a debug message indicating what went wrong.

Once again, I'd like to thank you for your work so far!

CMakeLists.txt
```diff
@@ -44,6 +44,7 @@ option(USE_HIDAPI        "Use hidapi as the HID backend"           OFF)
 option(USE_PCSC          "Enable experimental PCSC support"        ON)
 option(USE_WINHELLO      "Abstract Windows Hello as a FIDO device" ON)
 option(NFC_LINUX         "Enable NFC support on Linux"             ON)
+option(BLUETOOTH_LINUX   "Enable Bluetooth support on Linux"       ON)
```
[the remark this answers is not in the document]
akemnade: yes, makes sense

CMakeLists.txt
```diff
@@ -285,6 +288,12 @@ else()
 		add_definitions(-DUSE_NFC)
 	endif()
 
+	if(BLUETOOTH_LINUX)
+		add_definitions(-DUSE_BLUETOOTH)
+		pkg_search_module(SYSTEMD libsystemd REQUIRED)
+		set(BLUETOOTH_LIBRARIES ${SYSTEMD_LIBRARIES})
```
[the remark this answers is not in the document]
akemnade: hmm, not every platform has the systemd dependency.  If we want to put everything linux-specific into this if() then we need BLUETOOTH_LIBRARIES

e.g. macOS support which is in preparation at my branch mac-wip. There we do not have systemd, so I have
```
+       if(BLUETOOTH_MACOS)
+               add_definitions(-DUSE_BLUETOOTH)
+               set(BLUETOOTH_LIBRARIES "-framework Foundation -framework CoreBluetooth")
+       endif()
+
```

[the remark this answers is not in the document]
LDVG: OK. Would
```
pkg_search_module(BLUETOOTH libsystemd REQUIRED)
```
make more sense then?

CMakeLists.txt
```diff
@@ -399,6 +408,7 @@ include_directories(${ZLIB_INCLUDE_DIRS})
 link_directories(${CBOR_LIBRARY_DIRS})
 link_directories(${CRYPTO_LIBRARY_DIRS})
 link_directories(${HIDAPI_LIBRARY_DIRS})
+link_directories(${BLUETOOTH_LIBRARY_DIRS})
```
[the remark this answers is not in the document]
akemnade: or set _*DIRS above... see my other comment.

src/bluetooth_linux.c
```diff
@@ -0,0 +1,466 @@
...
+fido_bluetooth_read(void *handle, unsigned char *buf, size_t len, int ms)
+{
+	struct ble *dev = (struct ble *)handle;
+	ssize_t r;
+	if (fido_hid_unix_wait(dev->status_fd, ms, NULL) < 0)
+		return -1;
+
+	r = read(dev->status_fd, buf, len);
+	if ((size_t)r != len)
+		return -1;
```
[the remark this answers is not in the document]
akemnade: hmm, status_fd is of type SOCK_SEQPACKET, so there should be no half packets. The only possibility would be if the authenticator sends intentionally small packets. Does not make sense. As far as I understand, the spec does not explicitely allow or forbid that.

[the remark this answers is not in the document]
LDVG: An example of the behavior I'm seeing:

When creating a new credential through our `cred` example, the first message received back from my authenticator is 
```
82 00 01 02
```
i.e. a `KEEPALIVE` signaling `UP_NEEDED`.

However, the RX function is called with `len=20` (the control point size).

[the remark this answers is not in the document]
akemnade: ran into other problems while testing with control point length > 200 again.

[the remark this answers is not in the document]
akemnade: but before doing the last push I tested fido-token -L,  -I, ssh-keygen and ssh-login again with both authenticators.

src/ble.c
```diff
@@ -0,0 +1,253 @@
+#include "fido.h"
```
LDVG, line 1: These newly introduced files are missing a licensing header and copyright information. :-)

akemnade:
```
CMakeFiles/fido2_shared.dir/ble_linux.c.o.d -o CMakeFiles/fido2_shared.dir/ble_linux.c.o -c /__w/libfido2/libfido2/src/ble_linux.c
327
In file included from /__w/libfido2/libfido2/src/ble_linux.c:2:
328
/usr/include/elogind/systemd/sd-bus.h:96:43: error: ISO C restricts enumerator values to range of 'int' [-Wpedantic]
329
   96 |         SD_BUS_CREDS_UNIQUE_NAME        = 1ULL << 31,
330
      |                                           ^~~~
331
make[2]: *** [src/CMakeFiles/fido2.dir/build.make:569: src/CMakeFiles/fido2.dir/ble_linux.c.o] Error 1
```
seems like a problem in alpine + libelogind.

src/ble_linux.c
```diff
@@ -0,0 +1,497 @@
...
+	sd_bus_message_exit_container(reply);  /* {sv} */
+	return connected && resolved && has_service && paired;
+}
+
+static int
+init_ble_fido_dev(fido_dev_info_t *di,
+    const char *path, const char *name)
+{
+	memset(di, 0, sizeof(*di));
+	if (asprintf(&di->path, "%s%s", FIDO_BLE_PREFIX, path) &&
```
LDVG, line 417: Per my copy of `asprintf(3)`: 

> If memory  allocation wasn't possible, or some other error occurs, these functions will return -1, and the contents of strp are undefined.

i.e. if `asprintf` fails, calling `free(di->path)` is undefined behavior. We'll need something like:

```c
if (asprintf(&di->path, "%s%s", FIDO_BLE_PREFIX, path) == -1) {
        fido_log_debug("%s: asprintf", __func__);
        di->path = NULL;
        goto fail;
}

/* <snip> */

fail:
      free(di->path);
      /* more error handling */
```

LDVG (commented):
Thank you for the updates. I've added a set of new comments, and also left a set of (potentially?) unresolved comments from the previous round.

src/ble_linux.c
```diff
@@ -0,0 +1,497 @@
...
+			goto out;
+
+		if (sd_bus_call_method(dev->bus, "org.bluez", dev->paths.service_revision,
+				       DBUS_CHAR_IFACE, "WriteValue", NULL, NULL, "aya{sv}", 1, 0x20, 0) < 0)
+			goto out;
+
+		dev->controlpoint_size = ((size_t)cp_len[0] << 8) + cp_len[1];
+		if (sd_bus_call_method(dev->bus, "org.bluez", dev->paths.status,
+				       DBUS_CHAR_IFACE, "AcquireNotify", NULL, &reply, "a{sv}", 0) < 0)
+			goto out;
```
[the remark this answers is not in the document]
LDVG: Hm, looks like `acquire_status()` still leaks the `sd_bus_message`.

akemnade:
opened an issue for the alpine trouble:
https://gitlab.alpinelinux.org/alpine/aports/-/issues/15102

akemnade:
A summary of the alpine problem: 
later versions of sd-bus.h in systemd/elogind use __extension__ keyword to allow 64bit enums even in pedantic mode. So this is solved in alpine edge already and next release in November. Without __extension__ these things are tolerated in system includes (not something added via -I). E.g. debian does not add -Isomething in pkg-config --cflags for libsystemd, so that gets ignored, alpine does, but has a copy of the header files in /usr/include/systemd besides of /usr/include/elogind, so things are found  in any case.

So what can we do now:
- just accept the alpine problem until it gets fixed
- remove addition of include path for systemd in CMakeLists.txt
- remove -pedantic
- any hacks in the alpine build process (like removing /usr/include/elogind in the  alpine container)

jo-bitsch:
I'm very interested in that functionality and would be willing to help fixing the remaining issues.

The alpine issue referenced above should be fixed in the current stable (as of the new release of alpine v.3.19 released last December 7, 2023). As the pipeline refers to latest ( https://github.com/Yubico/libfido2/blob/f03f4c4d9d25e932066bef5f26b8b82a686ccd33/.github/workflows/alpine_builds.yml#L20) this might just work.

Unfortunately, I can't inspect the fuzzing issues, as the log expired.

DavidZemon:
I forked @akemnade's repo and rebased it on top of Yubico's `main`, then installed on my local Ubuntu 24.04 system. I then tested it with one of Token's devices (a Token BioStick - https://tokenring.com) and **IT WORKED**! Well done @akemnade.

The UX isn't where I'd like it just yet. Our device is very small and does not remain on and connected at all times. So what I need is for this BLE feature to do a quick scan of nearby BLE devices, check if any of those devices are advertising the FIDO service UUID (`0xFFFD`), and if so, attempt to connect if not already connected. I don't know if I'll be given permission to work on this as part of work hours, but will definitely keep you updated if I'm able to put any time toward it.

akemnade:
I am still interested in having this merged. I am a bit puzzled about how to proceed with this fuzzy test issues. As I understand correctly, there are no actual problems seen with this code but fuzzing has to be set up correcly.

What I am wondering about is: To me the idea would be to basically bombard the code with random dbus input and look if something explodes. To achieve that I would think that functions like sd_bus_call_method() would have to be mocked. But then there are other functions like sd_bus_message_read_basic() which are just processing such input further. To me it feels a bit strange if such functions have to be mocked, too.

Regarding UX: I think the a bluetooth gatt profile has to be registered, so auto-connection is enabled to paired devices having that feature. I have some experience in doing such stuff, so I would do that, if the basic stuff would get in.

BTW: there is some mac-wip branch providing bluetooth support for libfido2 on macOS.

PR #704, opened 2023-06-20
assert: introduce `fido_assert_authdata_raw_{ptr,len}()`

### LINKED ISSUE #690: Expose raw Authenticator Data bytes to library users
The current interface for assertions has two fields returning [Authenticator Data](https://www.w3.org/TR/webauthn-2/#sctn-authenticator-data):
https://github.com/Yubico/libfido2/blob/f1a74c8f2792e269a0c47f156e13e6ddb457b1d3/src/fido/types.h#L201-L202

* `fido_authdata_t` is parsed authenticator data (not fully: [some fields are missing](https://github.com/Yubico/libfido2/blob/f1a74c8f2792e269a0c47f156e13e6ddb457b1d3/src/fido/types.h#L115-L121) -- but that's not the point of this issue!)
* `fido_blob_t` is a CBOR-encoded blob. When decoding this blob, I found that it's simply the raw authenticator data bytes wrapped in a CBOR structure. For example, the library returns `58250FEE26D2CB710296BE673B8F4688857064EFFEEBB68A67452649B1BA0A5DCB690100000027`. The byte (0x58) is "byte string" (see [this](https://www.rfc-editor.org/rfc/rfc8949.html#name-indefinite-length-byte-stri)), and the second (0x25) is the length (37 in decimal).

Why not return the raw authenticator data bytes? Would you accept a PR adding `fido_raw_authdata` to simplify the caller's job and avoid a CBOR decoding operation? Other libraries wrapping libfido2 are also waiting for this: 
https://github.com/keys-pub/go-libfido2/blob/2399f32e9077b2b1f651bb675505c8d6e654b833/fido2.go#L153-L161

LDVG:
Resolves #690.

### DIFF
--- a/fuzz/export.gnu
+++ b/fuzz/export.gnu
@@ -20,6 +20,8 @@
 		fido_assert_allow_cred;
 		fido_assert_authdata_len;
 		fido_assert_authdata_ptr;
+		fido_assert_authdata_raw_len;
+		fido_assert_authdata_raw_ptr;
 		fido_assert_blob_len;
 		fido_assert_blob_ptr;
 		fido_assert_clientdata_hash_len;
--- a/fuzz/fuzz_assert.c
+++ b/fuzz/fuzz_assert.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019-2022 Yubico AB. All rights reserved.
+ * Copyright (c) 2019-2023 Yubico AB. All rights reserved.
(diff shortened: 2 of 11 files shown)


LDVG:
investigating an approach more similar to `cred.c`, please hold

martelletto:
LGTM

kongeo (approved):

PR #706, opened 2023-06-23
Improve OpenSSL detection & document requirements

ngie-eign:
The primary goal of this change is to incorporate the CMake module, `FindOpenSSL`, to allow end-users to better configure which version of OpenSSL they wish to target when building libfido2.

This allows end-users to build on systems with a version of OpenSSL that does not match the default version, e.g., with 3.0.9 on a system that comes with 1.1t by default (FreeBSD 13.2-RELEASE).

While here, document some unmentioned requirements, i.e.,

- cmake is required (after this change the minimum version is 3.18).
- pcsclite is required for smart card support.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,7 +7,8 @@
 cmake_policy(SET CMP0025 NEW)
 
 project(libfido2 C)
-cmake_minimum_required(VERSION 3.0)
+# 3.18+ supports OpenSSL 3 detection with FindOpenSSL.
+cmake_minimum_required(VERSION 3.18)
 # Set PIE flags for POSITION_INDEPENDENT_CODE targets, added in CMake 3.14.
 if(POLICY CMP0083)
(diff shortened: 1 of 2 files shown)


LDVG:
Hi,

For my own understanding, is setting `PKG_CONFIG_PATH` not a sufficient mechanism to achieve the same thing?

Thank you.

ngie-eign:
> Hi,
> 
> For my own understanding: is setting `PKG_CONFIG_PATH`, which we already support, not a sufficient mechanism to achieve the same thing?
> 
> Thank you.

Hi @LDVG ! That might work -- I'll need to doublecheck if it does (sometime next week).

LDVG:
Any updates on this issue?  See also #718 (resolved by #719) which ultimately results in CMake also appending `CMAKE_PREFIX_PATH`, `CMAKE_FRAMEWORK_PATH`, and `CMAKE_APPBUNDLE_PATH` to the pkg-config search path.

PR #707, opened 2023-06-28
tools: add FLAG_CD

LDVG:
A flag used to interpret the first line of input as raw client data rather than a hash of the client data. This is useful for `fido2-cred -M` and `fido2-assert -G` under the Windows Hello backend.


martelletto:
LGTM

kongeo (approved):

PR #708, opened 2023-06-29
build: unbreak builds with _FORTIFY_SOURCE=1

ynezz:
Some downstream build systems allow passing custom _FORTIFY_SOURCE values and this results in conflicting compiler flags and build failures:

 <command-line>: error: "_FORTIFY_SOURCE" redefined

So lets fix it by undefining any previous _FORTIFY_SOURCE values and then forcing _FORTIFY_SOURCE=2.


ynezz:
Example of such failure https://github.com/openwrt/packages/actions/runs/5410066043/jobs/9831013719

LDVG:
Hm, wouldn't it be preferred to honor the environment's wishes for a specific `_FORTIFY_SOURCE` level?

ynezz:
> Hm, wouldn't it be preferred to honor the environment's wishes for a specific `_FORTIFY_SOURCE` level?

I've looked at fadcc9847b13 introducing this option and have assumed, that its desired. Would for example `_FORTIFY_SOURCE=3` work out of the box? Either way, I'm find adjusting it as desired.

LDVG:
> I've looked at https://github.com/Yubico/libfido2/commit/fadcc9847b137b2ddff64e7853c1c83ec2152806 introducing this option and have assumed, that its desired.

It should be seen as our desired default for release builds. But I'd argue that if `-D_FORTIFY_SOURCE` is set by the caller, we should assume that they know what they're doing respect its value.

> Would for example _FORTIFY_SOURCE=3 work out of the box?

I've only tested it briefly;  it certainly _should_ work. If it doesn't, then that's probably a bug.

ynezz:
> we should assume that they know what they're doing respect its value.

Ok, I agree and I'll rework it.

PR #716, opened 2023-08-12
Add PC/SC support on Linux

mlpformat:
This may allow additional NFC hardware to be used on Linux and others. I'm unable to test with NFC, but it works with a Cryptnox FIDO Card in a contact reader.

### DIFF
--- a/src/pcsc.c
+++ b/src/pcsc.c
@@ -6,12 +6,12 @@
  * SPDX-License-Identifier: BSD-2-Clause
  */
 
-#if __APPLE__
+#if __APPLE__ || __UNIX__
 #include <PCSC/wintypes.h>
 #include <PCSC/winscard.h>
 #else
 #include <winscard.h>
-#endif /* __APPLE__ */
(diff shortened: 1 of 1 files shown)


LDVG:
Hi,

We rely on `pkg-config` to set the appropriate `-I` flag. See e.g. our `cmake` output

```
<snip>
-- PCSC_INCLUDE_DIRS: /usr/include/PCSC
-- PCSC_LIBRARIES: pcsclite
-- PCSC_LIBRARY_DIRS: /usr/lib
-- PCSC_VERSION: 1.9.9
<snip>
```

and the output of a call to `make`

```
[ 15%] Building C object src/CMakeFiles/fido2_shared.dir/pcsc.c.o
cd /__w/libfido2/libfido2/build-Debug/src && /usr/bin/gcc <snip> -I/usr/include/PCSC <snip> 
```

(additional compiler flags omitted for brevity)

The current choice of using the `#else` branch for Linux should therefore be correct and I'll be closing this with no action. Let us know if there's something we're missing.

Thanks!

PR #735, opened 2023-10-16
tools: add CLI flag to force FIDO2

TheOneric:
Matching the exisitng `-u` flag to force U2F.
This can be convenient for checking that a device works correctly under FIDO2 and does not rely on fallback to U2F.

Specifically I’m currently investigating a device which should support FIDO2, failing under the CTAP2/FIDO2 mode of another implementation and wanted to cross-check that the device actually works with CTAP2 in libfido2 *(it does)*. I thought this may also be helpful to other so here’s the patch I used including documentation updates.

### DIFF
--- a/man/fido2-assert.1
+++ b/man/fido2-assert.1
@@ -170,6 +170,12 @@ By default,
 .Nm
 will use FIDO2 if supported by the authenticator, and fallback to
 U2F otherwise.
+.It Fl 2
+Obtain an assertion using only FIDO2.
+By default,
+.Nm
+will use FIDO2 if supported by the authenticator, and fallback to
+U2F otherwise.
 .It Fl v
 If obtaining an assertion, prompt the user for a PIN and request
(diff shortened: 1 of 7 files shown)


LDVG:
Hi,

While we can see how it could be useful for debugging your own implementation, we are not necessarily sure that the tools provided by libfido2 should learn it.  After all, they default to FIDO2 and drops to U2F only if the authenticator does not implement the former.  

Debug output can be used to [clearly see when U2F fallback happens](https://github.com/Yubico/libfido2/blob/92e3d895c30a4a1a92ecaa698ee19e08aa06c4a1/src/dev.c#L202).  The output from `fido2-cred -M` should also be indicative (attestation format "fido-u2f", for example).  Should you absolutely need to force FIDO2 behavior, that is best done in the implementation you're testing or by modifying one of our examples to do so.

Thank you for sharing the patch!

PR #738, opened 2023-10-31
Add option to allow building with BoringSSL

nacho:
To build with BoringSSL CRYPTO_INCLUDE_DIRS
and CRYPTO_LIBRARY_DIRS must be defined

Also fix all the compile issues when building
with BoringSSL

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -44,6 +44,7 @@ option(USE_HIDAPI        "Use hidapi as the HID backend"           OFF)
 option(USE_PCSC          "Enable experimental PCSC support"        ON)
 option(USE_WINHELLO      "Abstract Windows Hello as a FIDO device" ON)
 option(NFC_LINUX         "Enable NFC support on Linux"             ON)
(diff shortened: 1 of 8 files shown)


nacho:
I went at the end for a more conservative approach given the CI failures. We can revisit it later on. Can you please approve again the run? Thanks

LDVG:
Hi,

Thank you for your patch.  While we're generally not opposed to an idea like this, maintaining support for BoringSSL which [has no guarantees of API stability](https://github.com/google/boringssl/blob/f714cb2935906a2f085c3b89d7e206af94627b56/README.md?plain=1#L5-L8) is a tough ask. It might make more sense for us to refactor libfido2 to use a thin crypto abstraction layer that is statically linked. That would make it easier for a third party to write their own wrapper to use something else if they'd prefer. This has been [discussed previously in the context of other crypto libraries](https://github.com/Yubico/libfido2/discussions/681).

nacho:
Sounds fair, thanks for the comment

PR #746, opened 2023-11-15
Install pkgconfig files also on msvc

nacho:


nacho:
These days it is perfectly normal that you might want to build with msvc using meson on windows. By installing the pc file also on windows makes it a lot simpler to build against fido2.

nacho:
That is correct, meson does the translation without any issues.

martelletto:
LGTM

LDVG (approved):
Thanks for the clarification!

PR #766, opened 2023-12-29
Change URL for LibreSSL dependency to avoid issues with the windows build

bobomb:
Seems like the powershell `Invoke-WebRequest` command to download LibreSSL is failing - I suspect the URL being used is being blocked, as the URL works fine when opened from a browser, but does not when using Invoke-WebRequest.

I was able to workaround this by changing the URL in windows\const.ps1 from `https://cloudflare.cdn.openbsd.org/pub/OpenBSD/LibreSSL` to `https://ftp.openbsd.org/pub/OpenBSD/LibreSSL/`, which works with `Invoke-WebRequest`

### DIFF
--- a/windows/const.ps1
+++ b/windows/const.ps1
@@ -5,7 +5,7 @@
 
 # LibreSSL coordinates.
 New-Variable -Name 'LIBRESSL_URL' `
-    -Value 'https://cloudflare.cdn.openbsd.org/pub/OpenBSD/LibreSSL' `
+    -Value 'https://ftp.openbsd.org/pub/OpenBSD/LibreSSL' `
     -Option Constant
 New-Variable -Name 'LIBRESSL' -Value 'libressl-3.8.2' -Option Constant
 New-Variable -Name 'CRYPTO_LIBRARIES' -Value 'crypto' -Option Constant


LDVG:
Hi,

Thanks for the report.  What version of Windows are you running?  Does the alternate patch below help?

```diff
diff --git a/windows/build.ps1 b/windows/build.ps1
index 5ac045b3..c15f315d 100644
--- a/windows/build.ps1
+++ b/windows/build.ps1
@@ -17,6 +17,7 @@ param(
 
 $ErrorActionPreference = "Stop"
 [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
+[Net.ServicePointManager]::SecurityProtocol += [Net.SecurityProtocolType]::Tls13
 
 . "$PSScriptRoot\const.ps1"
 

```

LDVG:
It'd appear [Windows 10 does not support TLS 1.3](https://learn.microsoft.com/en-us/windows/win32/secauthn/protocols-in-tls-ssl--schannel-ssp-#tls-protocol-version-support), and the Cloudflare mirror only offers TLS 1.3 (as tested via `openssl s_client`). The alternate patch above does work fine on Windows 11 and Windows Server 2022; our release system and GitHub pipelines also seem to work fine OOB for some reason.

I have slightly reworded your patch to reflect this and to keep the commit title at a reasonable length, but its contents remain the same (barring the removal of the extra trailing forward slash). If this problem bubbles up again sometime we might have to try a different approach. :-)

Thank you!

LDVG (approved):

bobomb:
Whoops, didn't mean to accidently close this PR.

kongeo (approved):

PR #767, opened 2024-01-02
Adds max cred blob length to -I

WillSmartYubico:
Not sure if this information is useful, just added it for completeness.


LDVG (approved):

LDVG:
Thanks @WillSmartYubico!  (squashed & merged)

PR #769, opened 2024-01-10
hidraw: fallback to  HID_NAME

jo-bitsch:
if we can't get the vendor/product string of the USB attributes, fallback to the HID_NAME string collected from the uevent.

This will be useful in linux when you provide a virtual fido2 device via `/dev/uhid` that emulates a USB connected HID device.

### DIFF
--- a/src/hid_linux.c
+++ b/src/hid_linux.c
@@ -77,7 +77,7 @@ out:
 
 static int
 parse_uevent(const char *uevent, int *bus, int16_t *vendor_id,
-    int16_t *product_id)
+    int16_t *product_id, char **hid_name)
 {
 	char			*cp;
 	char			*p;
@@ -97,8 +97,12 @@ parse_uevent(const char *uevent, int *bus, int16_t *vendor_id,
 				*vendor_id = (int16_t)y;
 				*product_id = (int16_t)z;
 				ok = 0;
-				break;
 			}
+			continue;
+		}
+		if (strncmp(p, "HID_NAME=", 9) == 0 && hid_name != NULL && *hid_name == NULL) {
+			*hid_name = strdup(p + 9);
+			continue;
 		}
 	}
 
@@ -137,6 +141,7 @@ copy_info(fido_dev_info_t *di, struct udev *udev,
 	char			*uevent = NULL;
 	struct udev_device	*dev = NULL;
 	int			 bus = 0;
+	char		*hid_name = NULL;
 	int			 ok = -1;
 
 	memset(di, 0, sizeof(*di));
@@ -148,7 +153,7 @@ copy_info(fido_dev_info_t *di, struct udev *udev,
 		goto fail;
 
 	if ((uevent = get_parent_attr(dev, "hid", NULL, "uevent")) == NULL ||
-	    parse_uevent(uevent, &bus, &di->vendor_id, &di->product_id) < 0) {
+	    parse_uevent(uevent, &bus, &di->vendor_id, &di->product_id, &hid_name) < 0) {
 		fido_log_debug("%s: uevent", __func__);
 		goto fail;
 	}
@@ -165,6 +170,13 @@ copy_info(fido_dev_info_t *di, struct udev *udev,
 		di->manufacturer = strdup("");
 	if ((di->product = get_usb_attr(dev, "product")) == NULL)
 		di->product = strdup("");
+	if ( di->manufacturer != NULL && *di->manufacturer == '\0' && 
+		 di->product != NULL && *di->product == '\0' && 
+		 hid_name != NULL ) {
+		
+		free(di->product);
+		di->product = strdup(hid_name);
+	}
 	if (di->path == NULL || di->manufacturer == NULL || di->product == NULL)
 		goto fail;
 
@@ -174,6 +186,7 @@ fail:
 		udev_device_unref(dev);
 
 	free(uevent);
+	free(hid_name);
 
 	if (ok < 0) {
 		free(di->path);


jo-bitsch:
There was a off by one in the buffer size for the string, which the pipeline runner caught.  I updated the PR accordingly.

jo-bitsch:
There are 2 warning reported by the pipelines.

One is: 
```
/home/runner/work/libfido2/libfido2/src/hid_linux.c:177:23: warning: Null pointer passed to 1st parameter expecting 'nonnull' [core.NonNullParamChecker]
  177 |                 char* first_space = strchr(hid_name, ' ');
      |                                     ^~~~~~~~~~~~~~~~~~~~~

```
This should be resolved by adding a hid_name != NULL to the previous condition.


The other is:
```
/home/runner/work/libfido2/libfido2/src/hid_linux.c:94:14: warning: Potential memory leak [unix.Malloc]
   94 |         while ((p = strsep(&cp, "\n")) != NULL && *p != '\0') {
      |                     ^~~~~~
```

I didn't change this line in my patch though.

This is in the context of `s = cp = strdup(uevent)` where later `free(s);` is called. I'm therefore not sure if this is a false positive. I'll think a bit more about how to best solve this.

jo-bitsch:
My bad, it was in my patch after all, just the warning was slightly misleading to me. Once I ran
```
scan-build-17 --use-cc=clang-17 cmake -DCMAKE_BUILD_TYPE=Debug ..
```
and then
```
 scan-build-17 --use-cc=clang-17 --status-bugs make -C build-Debug
```
similar to how the pipeline ran the code, the potential memory leak became apparent.

I updated the PR accordingly.

jo-bitsch:
I really appreciate the static analysis in the pipeline.

jo-bitsch:
The fuzzer discovered another issue, which I think is a false positive. Nevertheless I slightly rewrote that line in copy_info to make it clear to the compiler that  `di->manufacturer` and `di->product` cannot be `NULL`.

src/hid_linux.c
```diff
@@ -97,8 +98,14 @@ parse_uevent(const char *uevent, int *bus, int16_t *vendor_id,
 				*vendor_id = (int16_t)y;
 				*product_id = (int16_t)z;
 				ok = 0;
-				break;
 			}
+			continue;
+		}
+		if (strncmp(p, "HID_NAME=", 9) == 0 && hid_name != NULL) {
+			if (sscanf(p + 9, "%79[^\n]s", name) == 1 && *hid_name == NULL) {
```
LDVG, line 105: Any particular reason we're not just calling `strdup(p + 9)`, and skipping the `sscanf()` parsing?

src/hid_linux.c
```diff
@@ -165,6 +173,26 @@ copy_info(fido_dev_info_t *di, struct udev *udev,
 		di->manufacturer = strdup("");
 	if ((di->product = get_usb_attr(dev, "product")) == NULL)
 		di->product = strdup("");
+	if ( di->manufacturer != NULL && *di->manufacturer == '\0' && 
+		 di->product != NULL && *di->product == '\0' && 
+		 hid_name != NULL ) {
+		char* first_space = strchr(hid_name, ' ');
```
LDVG, line 179: While I can understand the intention here, I'd probably be more in favor of not manipulating the string in any way and punting that potential work to the application.  This would simplify handling here: e.g. if `di->manufacturer` and `di->product` are both unset, set `di->product` to `hid_name`.

LDVG (commented):
Hi, thanks for your patch.   Am I correct in understanding that you mostly want to expose a friendly name of the (virtual) HID device as created by `UHID_CREATE2`, e.g. when listing authenticators connected to the system?

src/hid_linux.c
jo-bitsch: Yes, actually:
`cp` holds the complete rest of the string of `uevent`, which potentially includes all of the newlines and following lines. `sscanf` already properly deals with 0 terminator and newlines.

As sscanf was already part of the function, I was thinking, this would be fair game, as long as the buffer length and everything are set accordingly.

Alternative solutions could use `strtok` , `strsep` or searching for newline and then use `strncpy`.

Do you want me to change it?

src/hid_linux.c
jo-bitsch: I'm fine either way. Technically, just splitting at the first space is not guaranteed to be correct, as the manufacturer string might include spaces itself, e.g. `Yubico AB`. In practice, most companies try to make it a single word and leave out the ` AB` part in this string. I'll follow your guidance/stated preference.

src/hid_linux.c
LDVG: Hm, not sure I follow.  We're already using `strsep()` call used to iterate over each line of the `uevent`:

https://github.com/Yubico/libfido2/blob/0071ac0e8d738b728acd92e4f143834a0d2ac578/src/hid_linux.c#L94

This should already guarantee that `p` does not contain any `\n` and is be NUL-terminated.

jo-bitsch:
Yes, that is exactly the point: Make the listing of connected authenticators more human-friendly for virtual hid device (using `/dev/uhid`) or non-USB connected HID devices (if compiled with `FIDO_HID_ANY`)

jo-bitsch:
I updated the PR to follow your 2 recommendations.

LDVG:
Paging in @martelletto: I see that something similar to this has both been [added to](https://github.com/Yubico/libfido2/commit/60992d510dc0319b3366ad7f44f875aca7658528#diff-10af2e83f2ada0a5e41a2b34a1a8b4fc889b4b7c768254c47fe883fd40613e13R226) and [removed from](https://github.com/Yubico/libfido2/commit/e11d4dab44488f84f3160b8ef58af4ee895c4d2f#diff-10af2e83f2ada0a5e41a2b34a1a8b4fc889b4b7c768254c47fe883fd40613e13L221) libfido2 previously. Do you have any additional context?

LDVG:
We merged your changes with some additional tweaks on top. Thank you for your contribution.

Issue #749, opened 2023-11-20
x5c data in attestation statement


yesvivek:
I am trying to implement Relying Party using libfido2 on linux environment. All the client interactions are done by browser using webauth JS APIs. 

According to webauthn spec I understand the _x5c_ under _attStmt_ can be of array type. But libfido2 stores only the leaf certificate and ignores the other certificates. Is it possible to update libfido2 to provide the other certificates too? This will prevent the callers from parsing the attestation object again. Maybe expose another API to accomplish this?

yesvivek:
When using webauth JS API, _PublicKeyCredential.AuthenticatorAttestationResponse.attestationObject_ gives a cbor data containing attestation, format, authData. Can libfido2 expose an API to consume this cbor-data (unparsed, as-is) directly? This will make the library more easy going with webauthn JS APIs. Without this the caller has to do cbor_load on attestationObject, find attStmt, then serialize it so it can be used with libfido2.

Or, if callers should parse this _attestationObject_, can libfido2 take attestation statement in libcbor structure? Because caller is already parsing the _attestationObject_ to extract the statement, why not just pass the libcbor structure to it? If you dont want to tie libfido2 to libcbor, probably taking the raw attestationObject might be better I feel.

yesvivek:
By taking the _PublicKeyCredential.AuthenticatorAttestationResponse.attestationObject_ data encoded in base64url and providing an API to get the x5c array, RP doesn't have to deal with cbor at all, just like how it is with libraries in other languages like NodeJS/Python.

LDVG:
Hi,

> According to webauthn spec I understand the x5c under attStmt can be of array type. But libfido2 stores only the leaf certificate and ignores the other certificates. Is it possible to update libfido2 to provide the other certificates too? This will prevent the callers from parsing the attestation object again. Maybe expose another API to accomplish this?

Out of curiosity, which authenticator are you using that is returning a certificate chain? Given a use case, we could certainly try to accommodate it.

> When using webauth JS API, PublicKeyCredential.AuthenticatorAttestationResponse.attestationObject gives a cbor data containing attestation, format, authData. Can libfido2 expose an API to consume this cbor-data (unparsed, base64url-encoded) directly? This will make the library more easy going with webauthn JS APIs. Without this the caller has to do cbor_load on attestationObject, find attStmt, then serialize it so it can be used with libfido2.

[...]

PR #782, opened 2024-02-07
Set attestation object containing CBOR encoded authData, fmt and attStmt

yesvivek:
Applications using Javascript WebAuthn APIs to communicate with FIDO2 devices get CBOR encoded attestation object containing authData, fmt and attStmt; more details about the JS API are given at [MDN](https://developer.mozilla.org/en-US/docs/Web/API/AuthenticatorAttestationResponse/attestationObject). Since there are no straight ways to set this value, I have updated libfido to set internal values after parsing the CBOR encoded attestation-object.
This will mean that the applications using libfido doesn't have to handle CBOR data at all, just like how other high level libraries in other languages behave.
Summary of changes:
  - `fido_cred_set_attobj` API is added to set attestation object containing CBOR encoded authData, fmt and attStmt
  - Supporting routines to parse and set the internal structure.
 
Relevant issue #749 .


src/cred.c
```diff
@@ -736,6 +736,88 @@ fido_cred_set_attstmt(fido_cred_t *cred, const unsigned char *ptr, size_t len)
 	return (r);
 }
 
+static int
+parse_attobj(const cbor_item_t *key, const cbor_item_t *val, void *arg)
```
LDVG, line 740: We have a very similar function in `winhello.c`,  see `decode_attobj()`.  If we were to introduce this public API, I feel like it'd make sense to consolidate them.  

One approach would be to move `winhello.c`'s `decode_attobj()` to e.g. `cbor.c` and introduce a `cbor_decode_attobj()` function, like you've done (but I don't think we need it to take a function pointer). Then call this function from both `winhello.c/translate_winhello_cred()` and the new `fido_cred_set_attobj()`.

To ensure a consistent state, `fido_cred_set_attobj()` would have to call `fido_cred_reset_rx()` first.

Any thoughts?

yesvivek: Did not know that winhello had same function, so ended up adding a new one; fixed now.
`fido_cred_reset_rx` was tempting but it clears large-blob too and I don't know what it does. Anyways, now using `fido_cred_reset_rx`.
Pushed new changes according to your suggestion.

yesvivek: Ok, in that case I will add `fido_cred_clean_attobj()`. 
Should `winhello.c/translate_winhello_cred()` also be updated to use it? @LDVG

LDVG: That should not be necessary, `fido_winhello_make_cred()` already calls `fido_cred_reset_rx()` before calling `translate_winhello_cred()`

yesvivek: Ok, pushed a new change to add `fido_cred_clean_attobj()`.

src/cred.c
```diff
@@ -549,6 +549,15 @@ fido_cred_clean_attstmt(fido_attstmt_t *attstmt)
 	memset(attstmt, 0, sizeof(*attstmt));
 }
 
+static void
+fido_cred_clean_attobj(fido_cred_t *cred)
```
LDVG, line 553: Would it make sense for `fido_cred_reset_rx()` to call this function?

yesvivek: Yes, that makes sense. Pushed it.

yesvivek:
Assuming this change will be merged, any idea when the next release is?

LDVG (commented):
I pushed a few additional tweaks on top of your changes, let me know if anything looks off. :slightly_smiling_face: 

> Assuming this change will be merged, any idea when the next release is?

There's no exact dates planned yet. It's likely it'd be sometime this spring.

man/fido_cred_set_authdata.3
```diff
@@ -157,6 +161,24 @@ The latter two are meant to be used in contexts where the
 credential's complete attestation statement is not available or
 required.
 .Pp
+The attestation object passed to
+.Fn fido_cred_set_attobj
+must be a CBOR-encoded map containing
+.Fa authData ,
```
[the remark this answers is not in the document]
yesvivek: You are correct, `.Dq` is appropriate here. Pushed the chage.

man/fido_cred_set_authdata.3
```diff
@@ -157,6 +161,24 @@ The latter two are meant to be used in contexts where the
...
+An application calling
+.Fn fido_cred_set_attobj
+does not need to call
+.Fn fido_cred_set_attstmt
+or
+.Fn fido_cred_set_fmt
+or
+.Fn fido_cred_set_authdata 
+or
+.Fn fido_cred_set_authdata_raw .
```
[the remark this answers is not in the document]
yesvivek: This is much better, pushed the change.

yesvivek:
@LDVG is there anything needed that is blocking merge with main?

LDVG:
> @LDVG is there anything needed that is blocking merge with main?

Apologies for the delays.  There shouldn't be much more needed to get this through, though a final review may take a little bit longer.  I squashed some of the intermediate commits, I hope that is fine with you.

LDVG:
Note (mostly to myself): Pushed a tentative fuzzer harness; needs seed corpora.

fuzz/fuzz_attobj.c
martelletto, line 1: missing copyright blurb

man/fido_cred_set_authdata.3
```diff
@@ -157,6 +161,22 @@ The latter two are meant to be used in contexts where the
 credential's complete attestation statement is not available or
 required.
 .Pp
+The attestation object passed to
```
martelletto, line 164: it might make sense (at a later point) to add a line or two about the purpose/utility of set_attobj

src/cbor.c
```diff
@@ -1132,6 +1132,64 @@ decode_attcred(const unsigned char **buf, size_t *len, int cose_alg,
...
+		goto fail;
+	}
+
+	if (!strcmp(name, "fmt")) {
+		if (cbor_decode_fmt(val, &cred->fmt) < 0) {
+			fido_log_debug("%s: cbor_decode_fmt", __func__);
+			goto fail;
+		}
+	} else if (!strcmp(name, "attStmt")) {
+		if (cbor_decode_attstmt(val, &cred->attstmt) < 0) {
```
martelletto, line 1154: something we need to reevaluate at one point is how we deal with duplicate entries in maps, which IIRC are allowed in CTAP CBOR. in some places (like cbor_decode_fmt) we error out, in others (cbor_decode_attstmt, alg case) we silently overwrite previously observed values

martelletto (commented):
left some comments; nothing major; LGTM

src/cbor.c
LDVG: Mainly as a reference to myself:

> something we need to reevaluate at one point is how we deal with duplicate entries in maps, which IIRC are allowed in CTAP CBOR

[§ 8. Message Encoding](https://fidoalliance.org/specs/fido-v2.1-ps-20210615/fido-client-to-authenticator-protocol-v2.1-ps-20210615.html#message-encoding) mentions

> All decoders SHOULD reject CBOR that is not validly encoded in the [CTAP2 canonical CBOR encoding form](https://fidoalliance.org/specs/fido-v2.1-ps-20210615/fido-client-to-authenticator-protocol-v2.1-ps-20210615.html#ctap2-canonical-cbor-encoding-form) and SHOULD reject messages with duplicate map keys.

> in others (cbor_decode_attstmt, alg case) we silently overwrite previously observed values

Yes, `alg` is the odd one out there. We could certainly add a demanding `COSE_UNSPEC`.

man/fido_cred_set_authdata.3
LDVG: I'll add this to my TODOs.

kongeo (approved):
LGTM!

PR #795, opened 2024-03-21
Fix potential memory leak

beldmit:
If first allocation succeeds and the 2nd fails, the `id` variable may be allocated and not freed. It's probably near to impossible case but it's worth fixing it for automation scan's sake


LDVG (approved):
Thanks!

kongeo (approved):

PR #796, opened 2024-03-22
Simplify getting an EVP_MD for all OpenSSL forks

botovq:
In all versions of OpenSSL, EVP_sha{1,256}() return a pointer to static read-only data. There is no need to copy that data into a buffer that is immediately freed after use.

In LibreSSL and OpenSSL 1.x, EVP_PKEY_CTX_set_signature_md() is a wrapper around EVP_PKEY_ctrl() whose void * argument isn't const (contrary to the documented signature), so we need to cast const away in some way. This preserves rs*_get_EVP_MD() to contain the pragma pollution necessary due to -Werror -Wcast-qual.

The negative performance impact of calling EVP_sha*() multiple times called out in the OpenSSL 3.x documentation should not apply in this code.

Closes: #794

### DIFF
--- a/src/rs1.c
+++ b/src/rs1.c
@@ -10,54 +10,14 @@
 
 #include "fido.h"
 
-#if defined(LIBRESSL_VERSION_NUMBER) && LIBRESSL_VERSION_NUMBER < 0x3050200fL
 static EVP_MD *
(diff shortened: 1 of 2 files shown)


LDVG (approved):
Thanks!

PR #810, opened 2024-06-05
man: clarify fido_cred_set_attobj purpose

LDVG:


kongeo (approved):

man/fido_cred_set_authdata.3
```diff
@@ -176,6 +176,9 @@ does not need to call
 .Fn fido_cred_set_authdata ,
 or
 .Fn fido_cred_set_authdata_raw .
+This function may be useful for applications interfacing with the
```
martelletto, line 179: `.Fn fido_cred_set_attobj` might be clearer and more succinct in this context. Perhaps also "useful in applications" or "useful to applications" instead of "for applications"?

LDVG: updated; thanks!

PR #812, opened 2024-06-15
shellcheck workflow + fixes

martelletto:
add a shellcheck workflow and implement required fixes. while here, add a couple more tests to `tools/test.sh`.

### DIFF
--- a/tools/include_check.sh
+++ b/tools/include_check.sh
@@ -1,15 +1,22 @@
-#!/bin/sh
+#!/bin/sh -u
 
 # Copyright (c) 2019 Yubico AB. All rights reserved.
 # Use of this source code is governed by a BSD-style
 # license that can be found in the LICENSE file.
 # SPDX-License-Identifier: BSD-2-Clause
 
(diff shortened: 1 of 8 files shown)


tools/include_check.sh
```diff
@@ -1,15 +1,22 @@
...
-			cc $CFLAGS -Isrc -xc -c - -o /dev/null 2>&1
-		echo "$f $CFLAGS $?"
+	try="cc $CFLAGS -Isrc -xc -c - -o /dev/null 2>&1"
+	git ls-files "$1" | grep '.*\.h$' | while read -r header; do
+		if echo "$header" | grep -Eq "$SKIP"; then
+			echo "Skipping $header"
+		else
+			body="#include \"$header\""
+			echo "echo $body | $try"
+			echo "#include \"$header\"" | eval "$try"
```
LDVG, line 18: Since we have it, would `echo $body` make sense here?

martelletto: Sure; addressed in new branch

PR #819, opened 2024-09-13
U2F key_lookup: do not fail upon SW_WRONG_LENGTH (0x6700)

James-ZHANG:
According to u2f spec: https://fidoalliance.org/specs/fido-u2f-v1.2-ps-20170411/fido-u2f-raw-message-formats-v1.2-ps-20170411.html#status-codes, U2F security keys can return `0x6700` in addition to `0x6A80` when given invalid key handles.

This occurs to our internally developed security keys when the key handle is too long.

### DIFF
--- a/src/fido/param.h
+++ b/src/fido/param.h
@@ -74,6 +74,7 @@
 
 /* ISO7816-4 status words. */
 #define SW1_MORE_DATA			0x61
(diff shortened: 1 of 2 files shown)


martelletto:
LGTM

LDVG (approved):
Thanks!

PR #824, opened 2024-10-02
cred: support for enterprise attestation

LDVG:
In short, the following user visible changes are applied:

- `fido_cred_set_entattest()` is added for the client to be able to request enterprise attestation;
- `fido_cred_entattest()` is added to query whether the authenticator performed enterprise attestation; and
- `fido2-cred -M -a` is added to request enterprise attestation (dito `examples/cred`).

Internally,

- `fido_dev_make_cred_tx()` learned how to encode the enterprise attestation parameter;
- `fido_dev_make_cred_rx()` learned how to decode the response; and
- likewise for the `winhello.c` equivalents.

Finally, the fuzzer and manual pages are updated accordingly.

Further resources:

- https://fidoalliance.org/specs/fido-v2.1-ps-20210615/fido-client-to-authenticator-protocol-v2.1-ps-20210615.html#sctn-feature-descriptions-enterp-attstn
- https://fidoalliance.org/specs/fido-v2.1-ps-20210615/fido-client-to-authenticator-protocol-v2.1-ps-20210615.html#op-makecred-step-if-enterpriseAttestation


martelletto:
we seem to be missing [MAN_ALIAS](https://github.com/Yubico/libfido2/blob/main/man/CMakeLists.txt#L47) bits for the new functions, otherwise LGTM. for my own education, what does `fido_cred_entattest()` signify? what is the application expected to do with that information?

martelletto:
understood; thank you. is there wire data we could use for regression tests?

LDVG:
> understood; thank you. is there wire data we could use for regression tests?

We could definitely record some. I'll make sure to do so before marking this PR as ready. Thanks!

LDVG:
I've added regression tests where we can assert that we parse the response to `fido_dev_make_cred()` correctly, this should now mostly be good to go.

kongeo (approved):
LGTM! :)

PR #826, opened 2024-10-11
Allow len < 16 for aes256-gcm encryption

James-ZHANG:
Only decryption should check `len < 16`.


LDVG:
Good catch.  As with your other PR, would you mind amending the commit message, excluding the trailing period? Looks good to me otherwise. Thank you!

James-ZHANG:
> As with your other PR, would you mind amending the commit message, excluding the trailing period?

Done.

LDVG (approved):

kongeo (approved):

PR #827, opened 2024-10-11
[examples/cred] Support self attestation

James-ZHANG:
Some attestation format, e.g. `packed`, allows self attestation [1]. And when that happens, `fido_cred_x5c_ptr` would be `NULL`, making `fido_cred_verify` fail.

[1] https://w3c.github.io/webauthn/#sctn-packed-attestation


LDVG:
Looks good to me!

One minor nitpick, could you please amend your commit message so that the subject/title line does not end in a period? Thank you!

James-ZHANG:
> could you please amend your commit message so that the subject/title line does not end in a period?

Done.

LDVG (approved):

PR #837, opened 2024-11-19
cred: support requesting multiple types

LDVG:


src/cred.c
```diff
@@ -1071,23 +1072,43 @@ fido_cred_set_fmt(fido_cred_t *cred, const char *fmt)
...
+
+	return (fido_cred_append_type(cred, cose_alg));
+}
+
 int
 fido_cred_type(const fido_cred_t *cred)
 {
-	return (cred->type);
+	if (cred->attcred.type != 0 || cred->type.len == 0)
+		return (cred->attcred.type);
```
martelletto, line 1109: this relies on `COSE_UNSPEC` being defined to zero; correct? (if `cred->type.len == 0`)

man/fido_cred_set_authdata.3
```diff
@@ -392,6 +394,13 @@ The type of a credential may only be set once.
 Note that not all authenticators support COSE_RS256, COSE_ES384, or
 COSE_EDDSA.
 .Pp
+The
+.Fn fido_cred_append_type
+function is the same as
+.Fn fido_cred_set_type
+but can be called repeatedly to request one of multiple types, in order
```
martelletto, line 401: consider rephrasing it as `Unlike fido_cred_set_type, fido_cred_append_type may be called repeatedly to request multiple different types` to avoid confusion. we may also want to point out (not necessarily here) that `fido_cred_set_authdata()`, `fido_cred_set_authdata_raw()`, and `fido_dev_make_cred()` may now change the credential type returned by `fido_cred_type()`.

martelletto:
left a couple of comments. looks good, although i'm not sure it justifies the additional complexity. in any case, we might want to add fuzzing bits and corresponding entries to the wiredata files so that the new code paths get properly exercised.

src/cred.c
LDVG: yes, we could certainly be more specific if that helps.

regress/cred.c
```diff
@@ -3021,6 +3021,36 @@ entattest(void)
...
 }
+static void
+multiple_cose(void)
+{
+	fido_cred_t *c;
+
+	c = alloc_cred();
+	assert(fido_cred_type(c) == COSE_UNSPEC);
+	assert(fido_cred_append_type(c, COSE_EDDSA) == FIDO_OK);
+	assert(fido_cred_type(c) == COSE_EDDSA); // compat: only algorithm req.
```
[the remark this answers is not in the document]
LDVG: thanks!

man/fido_cred_set_authdata.3
LDVG: yes, the part about where authdata updates the type may be surprising but at least it doesn't happen with `fido_cred_set_type()`, only `fido_cred_append_type()`.  however, the fact that the type changes may be enough reason to consider this breaking and warrant another function returning the `attcred.type` instead, but that may be equally confusing.

kathoum:
What is the plan about this patch? is it still considered a desirable feature, or has it been abandoned?

PR #846, opened 2024-12-06
cmake, add check on FUZZ_LDFLAGS

dacav:
Triviality, but it avoids triggering more "obscure" errors.

### DIFF
--- a/fuzz/CMakeLists.txt
+++ b/fuzz/CMakeLists.txt
@@ -18,6 +18,10 @@ set(FUZZ_LINKER_LANGUAGE "C" CACHE STRING "Linker language for fuzz harnesses")
 mark_as_advanced(FUZZ_LINKER_LANGUAGE)
 enable_language(${FUZZ_LINKER_LANGUAGE})
 
+if(NOT FUZZ_LDFLAGS)
+	message(FATAL_ERROR "Please define FUZZ_LDFLAGS")
+endif()
+
 # fuzz_cred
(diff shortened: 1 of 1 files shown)


fuzz/CMakeLists.txt
```diff
@@ -18,6 +18,10 @@ set(FUZZ_LINKER_LANGUAGE "C" CACHE STRING "Linker language for fuzz harnesses")
 mark_as_advanced(FUZZ_LINKER_LANGUAGE)
 enable_language(${FUZZ_LINKER_LANGUAGE})
 
+if(NOT DEFINED FUZZ_LDFLAGS)
```
LDVG, line 21: Would `if(NOT FUZZ_LDFLAGS)` make more sense here? I'm pretty sure an empty `FUZZ_LDFLAGS` will be problematic as well.

fuzz/CMakeLists.txt
```diff
@@ -18,6 +18,10 @@ set(FUZZ_LINKER_LANGUAGE "C" CACHE STRING "Linker language for fuzz harnesses")
...
+  message(FATAL_ERROR "Please define FUZZ_LDFLAGS")
```
LDVG, line 22: Inconsistent indentation, would you mind switching to a tab character?

LDVG (commented):
A few minor comments.

If we go for the below logic change, would `fuzz: assert non-empty FUZZ_LDFLAGS` (or `cmake: ...`) perhaps be a slightly more descriptive commit message?

LDVG (approved):

PR #857, opened 2025-03-03
Add support to pass a window handle to windows hello

Nehluxhes:
Add support to pass a specific window handle to winhello WebAuthNAuthenticatorGetAssertion and WebAuthNAuthenticatorMakeCredential

Makes for better parenting of different application windows instead of just relying on GetForegroundWindow/GetTopWindow (they are still used if not setting anything)

Note the handle is supposed to be a HWND (just a pointer) but I could not find a clean way to either define it (but then redefinition errors) or avoid the whole including windows.h/windef.h mess in fido.h, so void* it is

### DIFF
--- a/src/assert.c
+++ b/src/assert.c
@@ -802,6 +802,7 @@ fido_assert_reset_tx(fido_assert_t *assert)
 	memset(&assert->ext, 0, sizeof(assert->ext));
 	assert->rp_id = NULL;
 	assert->appid = NULL;
+	assert->window = NULL;
 	assert->up = FIDO_OPT_OMIT;
 	assert->uv = FIDO_OPT_OMIT;
 }
--- a/src/assert.c
+++ b/src/assert.c
@@ -664,12 +664,29 @@ fido_assert_set_winhello_appid(fido_assert_t *assert, const char *id)
 
 	return (FIDO_OK);
 }
+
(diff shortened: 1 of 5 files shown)


LDVG:
Thanks for the PR! :slightly_smiling_face: 

I've got two questions:

1. Would it make more sense to have this as an attribute on the device itself rather than the credential and assertion separately?
2. Did I understand it correctly that this essentially has no effect on resolving your issue?

Nehluxhes:
1. It could. Could I reuse the io_handle then ? I don't think you are using it for winhello and the HWND could be considered an input handle.
2. For now, unfortunately, no. If (when 🤞) microsoft fix it, it will though.

LDVG:
> It could. Could I reuse the io_handle then ? I don't think you are using it for winhello and the HWND could be considered an input handle.

Sounds like a reasonable take to me

> For now, unfortunately, no. If (when 🤞) microsoft fix it, it will though.

Is Microsoft aware of this issue?

Nehluxhes:
> Is Microsoft aware of this issue?

Yes. I sent a mail to the address from the winhello repo and got an answer it has been forwarded to more people.
So at the very least it has been read. Too bad there is no public bug tracker to follow though.

PR #876, opened 2025-06-02
check for allocations before freeing

Emmankoko:
in base64_encode(fido_cred_id_ptr(cred), fido_cred_id_len(cred),
	    &id) < 0 || base64_encode(fido_cred_user_id_ptr(cred),
	    fido_cred_user_id_len(cred), &user_id) < 0), 
either of id or user_id can come out of this call allocated.  non-zero return in base64_encode can also come from failed allocation of id and a successful allocation of user_id or the other way round. freeing both directly without checking if their allocation was successful risks freeing unallocated memory.

### DIFF
--- a/tools/credman.c
+++ b/tools/credman.c
@@ -140,8 +140,10 @@ print_rk(const fido_credman_rk_t *rk, size_t idx)
 
 	r = 0;
 out:
-	free(user_id);
-	free(id);
+	if (user_id)
+		free(user_id);
+	if (id)
+		free(id);
 
 	return r;
 }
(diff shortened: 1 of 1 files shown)


LDVG:
Hi,

There's no risk in passing a `NULL` pointer to `free(3)`, the C standard clearly states that no action occurs if the pointer is `NULL`.  See for example the the [latest publicly available version of the C99 standard ](https://open-std.org/jtc1/sc22/wg14/www/docs/n1256.pdf)section 7.20.3.2 *The `free` function*, paragraph 2 (emphasis mine) : 

> The `free` function causes the space pointed to by `ptr` to be deallocated, that is, made available for further allocation. **If `ptr` is a null pointer, no action occurs.**

Thank you for your efforts though!

Emmankoko:
right !

PR #884, opened 2025-07-21
io: fix for potentially unaligned access

safocl:
C standard:

> 6.3.2.3 Pointers 
> 7 A pointer to an object or incomplete type may be converted to a pointer to a different object or incomplete type. If the resulting pointer is not correctly aligned (note 57) for the pointed-to type, the behavior is undefined. Otherwise, when converted back again, the result shall compare equal to the original pointer. When a pointer to an object is converted to a pointer to a character type, the result points to the lowest addressed byte of the object. Successive increments of the result, up to the size of the object, yield pointers to the remaining bytes of the object.

> J.2 Undefined behavior
> 1  The behavior is undefined in the following circumstances:
> — Conversion between two pointer types produces a result that is incorrectly aligned
> (6.3.2.3).

as eximple [of code](https://github.com/Yubico/libfido2/blob/335e447778016410ac221c555a0faee2262e5798/src/io.c#L50):
```c
static int
tx_empty(fido_dev_t *d, uint8_t cmd, int *ms)
{
	struct frame	*fp;
	unsigned char	 pkt[sizeof(*fp) + 1];
	const size_t	 len = d->tx_len + 1;
	int		 n;

	memset(&pkt, 0, sizeof(pkt));
	fp = (struct frame *)(pkt + 1);
	fp->cid = d->cid;
	fp->body.init.cmd = CTAP_FRAME_INIT | cmd;

	if (len > sizeof(pkt) || (n = tx_pkt(d, pkt, len, ms)) < 0 ||
	    (size_t)n != len)
		return (-1);

	return (0);
}
```

`unsigned char pkt[sizeof(*fp) + 1];` and
`fp = (struct frame *)(pkt + 1);`
don't guarantee to be aligned as `struct frame` (as least the `struct frame` contains a `uint32_t` field).


LDVG:
We only support targets which work with the `packed` attribute. The issue you're referencing is not a problem for such targets if the only access to the struct member are as follows

```c
uint32_t x = fp->cid;
/* and */
fp->cid = x;
```

since the compiler knows `fp` is packed and must therefore emit instructions that take unaligned access into account. 
We would have an issue, even on targets that support the `packed` attribute, if we were accessing the struct member using

```c
uint32_t *y = &fp->cid;
*y = x;
```

since the compiler may then emit instructions that do not take unaligned access into account.

As I said, we may revisit this in the future in an attempt to make libfido2 more portable, but the behavior is well-defined for the platforms we support today.

PR #915, opened 2025-11-24
assert: return FIDO_ERR_PIN_REQUIRED when token returns FIDO_ERR_MISSING_PARAMETER and no pin was passed

bluca:
When a token is in FIPS mode the PIN is always required for certain operations, but the libfido2 returns a generic FIDO_ERR_MISSING_PARAMETER instead of a more useful parameter to the caller.

If a pin was not passed and we get FIDO_ERR_MISSING_PARAMETER, change the return code to FIDO_ERR_PIN_REQUIRED.

```
$ systemd-cryptenroll --fido2-device=auto --fido2-with-client-pin=no foo
<...>
libfido2: rx: payload_len=1
libfido2: fido_rx: buf=0x55c9ce170940, len=1
libfido2: 0000: 14
libfido2: cbor_parse_reply: blob[0]=0x14
libfido2: fido_dev_get_assert_rx: adjust_assert_count
Failed to ask token for assertion: FIDO_ERR_MISSING_PARAMETER
```

### DIFF
--- a/src/assert.c
+++ b/src/assert.c
@@ -340,12 +340,18 @@ fido_dev_get_assert(fido_dev_t *dev, fido_assert_t *assert, const char *pin)
 	}
 
 	r = fido_dev_get_assert_wait(dev, assert, pk, ecdh, pin, &ms);
-	if (r == FIDO_OK && (assert->ext.mask & FIDO_EXT_HMAC_SECRET))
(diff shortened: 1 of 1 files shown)


LDVG:
> So where would be the best place to do so?

I think the choice should be left for the application to decide what it'd like to do to in response to the particular error code.

bluca:
Is there a higher level API? I'd rather not muck about with sending/receiving command messages manually

PR #937, opened 2026-03-06
Ppuat 2

dacav:


src/bio.c
```diff
@@ -81,6 +103,7 @@ bio_tx(fido_dev_t *dev, uint8_t subcmd, cbor_item_t **sub_argv, size_t sub_argc,
 	memset(&f, 0, sizeof(f));
 	memset(&hmac, 0, sizeof(hmac));
 	memset(&argv, 0, sizeof(argv));
+	memset(&argv, 0, sizeof(our_token));
```
LDVG:
```suggestion
	memset(&our_token, 0, sizeof(our_token));
```

examples/puat-mcga.c
```diff
@@ -0,0 +1,309 @@
...
+ */
+
+#include <err.h>
+#include <sysexits.h>
+
+#include <fido.h>
+
+/* openssl sha256 -binary </dev/null | xxd -i */
+static const unsigned char CLIENT_DATA_HASH[] = {
+	0xe3, 0xb0, 0xc4, 0x42,
```
LDVG: mind reflowing these to e.g. 8 bytes per line?

src/bio.c
```diff
@@ -292,7 +308,7 @@ fido_bio_dev_get_template_array(fido_dev_t *dev, fido_bio_template_array_t *ta,
 {
 	int ms = dev->timeout_ms;
 
-	if (pin == NULL)
+	if (pin == NULL && fido_dev_puat_blob(dev) == NULL)
```
LDVG: as an aside, unrelated to these changes, given that [`uvBioEnroll` is a thing](https://fidoalliance.org/specs/fido-v2.2-ps-20250714/fido-client-to-authenticator-protocol-v2.2-ps-20250714.html#getinfo-uvbioenroll), a  FIXME should probably be to support that as well -- at that point, this entire check goes away

examples/puat-mcga.c
LDVG: This reads more like a `largeblob.c` example?   Same rationale as `paut-bio.c`.  🙂 


On our previous discussions about cleanup; what about having the MC parts of the example add the large blob, and the GA parts of the example remove the large blob ?  Covers more ground in terms of large blob API.

src/bio.c
```diff
@@ -279,7 +295,8 @@ bio_get_template_array_wait(fido_dev_t *dev, fido_bio_template_array_t *ta,
 {
 	int r;
 
-	if ((r = bio_tx(dev, CMD_ENUM, NULL, 0, pin, NULL, ms)) != FIDO_OK ||
+	if ((r = bio_tx(dev, CMD_ENUM, NULL, 0, pin, fido_dev_puat_blob(dev), ms))
+	    != FIDO_OK ||
```
LDVG: We do not break lines like this anywhere else. To be consistent with current style, we should use

```suggestion
	if ((r = bio_tx(dev, CMD_ENUM, NULL, 0, pin, fido_dev_puat_blob(dev),
	    ms)) != FIDO_OK ||
```

(and optionally squeeze the next statement in, but I'm less pedantic about that)

PR #962, opened 2026-04-01
Dacav/bio drop strict ifs

### LINKED ISSUE #960: Remove strict checks in bio
as an aside, unrelated to these changes, given that [`uvBioEnroll` is a thing](https://fidoalliance.org/specs/fido-v2.2-ps-20250714/fido-client-to-authenticator-protocol-v2.2-ps-20250714.html#getinfo-uvbioenroll), a  FIXME should probably be to support that as well -- at that point, this entire check goes away (same for similar cases here)

_Originally posted by @LDVG in https://github.com/Yubico/libfido2/pull/937#discussion_r2929630548_

dacav:
Fixes #960

### DIFF
--- a/src/bio.c
+++ b/src/bio.c
@@ -90,11 +90,25 @@ bio_get_uv_token(fido_dev_t *dev, const char *pin, uint8_t cmd,
 }
 
 static int
+resolve_token(fido_blob_t *token, fido_dev_t *dev, const char *pin, int *ms)
+{
+	const fido_blob_t *puat;
+
+	/* If available, prefer cached PUAT */
+	if ((puat = fido_dev_puat_blob(dev)) != NULL) {
+		if (fido_blob_set(token, puat->ptr, puat->len))
+			return FIDO_ERR_INTERNAL;
+		return FIDO_OK;
+	}
+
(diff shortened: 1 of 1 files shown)


src/bio.c
```diff
@@ -89,21 +89,30 @@ bio_get_uv_token(fido_dev_t *dev, const char *pin, uint8_t cmd,
...
 }
 
+static int
+resolve_token(fido_blob_t *token, fido_dev_t *dev, const char *pin, int *ms)
+{
+       const fido_blob_t *puat;
+
+       /* If available, prefer cached PUAT */
+       if ((puat = fido_dev_puat_blob(dev)) != NULL)
+               return fido_blob_set(token, puat->ptr, puat->len);
```
LDVG: We probably need to translate a potential `-1` return value here to `FIDO_ERR_INTERNAL`?

LDVG (commented):
Did we accidentally revert the removed PIN/token check for `fido_bio_dev_set_template_name()`? 🙂

src/bio.c
dacav: Ach, Pitfall!
I see that fido_blob_set() is violating the return value convention: should it return a boolean perhaps?
Thanks for spotting.

PR #964, opened 2026-04-02
man: expand documentation of PIN/UV Auth token

dacav:


man/fido_dev_get_puat.3
```diff
@@ -130,6 +130,25 @@ function returns a pointer to the PIN/UV Auth token currently associated with
 .Fa dev .
 The corresponding length can be obtained by calling
 .Fn fido_dev_puat_len .
+.Pp
+Associating a token with
+.Fa dev
+has the main benefit of reducing the user interaction to a minimum in
+a sequence of operations where each would require UV.
```
LDVG:
```suggestion
a sequence of operations where each would require some form of UV.
```

PR #979, opened 2026-05-21
fuzz: improvements

martelletto:
- soft reset wiredata in fuzz-{assert,cred};
- truncate hmac-secret blobs;
- fuzz_attobj: adjust cbor array allocation;
- fuzz_attobj: don't reset object on every mutation;
- fuzz_mgmt: mutate config_wire_data;
- trim unnecessary memset();
- wrap more symbols.


LDVG (approved):
LGTM, thanks!

PR #981, opened 2026-05-31
simplify and integrate tools/include_check.sh

martelletto:
- tools/include_check: simplify;
- actions/build-linux-gcc: run tools/include_check;
- src/extern.h: fix standalone inclusion;
- src/blob.h: include sys/types.h.

linux-only for now; other platforms require a config.h equivalent, which CMake does not provide


LDVG (approved):
LGTM

PR #987, opened 2026-06-16
pin: nullable (ecdh, pk) for fido_dev_get_uv_token()

LDVG:
An attempt to improve some of the PUAT handling, mainly in `assert.c` and `cred.c`.

`fido_dev_get_uv_token()` has historically always required the caller to provide `ecdh` and `pk`, though it's rarely needed for anything but the `hmac-secret` extension. By allowing these two arguments to be `NULL` and, if so, calling `fido_do_ecdh()` internally we can:

- remove `need_ecdh()` from both `cred.c` and `assert.c` and instead condition a call to `fido_do_ecdh()` on _only_ the presence of `hmac-secret`; and condition `fido_dev_get_uv_token()` _only_ on whether we're going to perform some form of UV;
- remove a few `xxx_get_uv_token()` helper functions that all do the very same thing as this new variant of `fido_dev_get_uv_token()`; and
- turn `cbor_add_uv_params()` into a pure encoder, narrowing its responsibilities (where it would previously also fetch a PIN/UV auth token from the device).

While here, adopt the changes from #962 to minimize conflicts.

Closes #962.

### DIFF
--- a/src/assert.c
+++ b/src/assert.c
@@ -83,7 +83,8 @@ parse_assert_reply(const cbor_item_t *key, const cbor_item_t *val, void *arg)
 
 static int
 fido_dev_get_assert_tx(fido_dev_t *dev, fido_assert_t *assert,
-    const es256_pk_t *pk, const fido_blob_t *ecdh, const char *pin, int *ms)
+    const es256_pk_t *pk, const fido_blob_t *ecdh, const fido_blob_t *token,
+    int *ms)
 {
 	fido_blob_t	 f;
 	fido_opt_t	 uv = assert->uv;
@@ -128,10 +129,9 @@ fido_dev_get_assert_tx(fido_dev_t *dev, fido_assert_t *assert,
 		}
 
 	/* user verification */
-	if (pin != NULL || fido_dev_puat_blob(dev) != NULL ||
-	    (uv == FIDO_OPT_TRUE && fido_dev_supports_permissions(dev))) {
-		if ((r = cbor_add_uv_params(dev, cmd, &assert->cdh, pk, ecdh,
-		    pin, assert->rp_id, &argv[5], &argv[6], ms)) != FIDO_OK) {
+	if (token != NULL) {
+		if ((r = cbor_add_uv_params(dev, &assert->cdh, token, &argv[5],
+		    &argv[6])) != FIDO_OK) {
 			fido_log_debug("%s: cbor_add_uv_params", __func__);
 			goto fail;
 		}
@@ -266,11 +266,12 @@ out:
 
 static int
 fido_dev_get_assert_wait(fido_dev_t *dev, fido_assert_t *assert,
-    const es256_pk_t *pk, const fido_blob_t *ecdh, const char *pin, int *ms)
+    const es256_pk_t *pk, const fido_blob_t *ecdh, const fido_blob_t *token,
+    int *ms)
 {
 	int r;
 
-	if ((r = fido_dev_get_assert_tx(dev, assert, pk, ecdh, pin,
+	if ((r = fido_dev_get_assert_tx(dev, assert, pk, ecdh, token,
 	    ms)) != FIDO_OK ||
 	    (r = fido_dev_get_assert_rx(dev, assert, ms)) != FIDO_OK)
 		return (r);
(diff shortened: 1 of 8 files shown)


src/assert.c
```diff
@@ -305,30 +306,17 @@ decrypt_hmac_secrets(const fido_dev_t *dev, fido_assert_t *assert,
...
-}
-
 int
 fido_dev_get_assert(fido_dev_t *dev, fido_assert_t *assert, const char *pin)
 {
-	fido_blob_t	*ecdh = NULL;
-	es256_pk_t	*pk = NULL;
-	int		 ms = dev->timeout_ms;
-	int		 r;
+	fido_blob_t		 token_store;
```
martelletto: perhaps s/token_store/throwaway_token/ or s/token_store/disposable_token/ or s/token_store/tmp_token/ for clarity?

src/config.c
```diff
@@ -44,15 +44,17 @@ static int
 config_tx(fido_dev_t *dev, uint8_t subcmd, cbor_item_t **paramv, size_t paramc,
     const char *pin, int *ms)
 {
+	const fido_blob_t *token = fido_dev_puat_blob(dev);
+	fido_blob_t token_store;
```
martelletto: same here

src/config.c
```diff
@@ -44,15 +44,17 @@ static int
 config_tx(fido_dev_t *dev, uint8_t subcmd, cbor_item_t **paramv, size_t paramc,
     const char *pin, int *ms)
 {
+	const fido_blob_t *token = fido_dev_puat_blob(dev);
```
martelletto: if possible, split the assignment from the declaration

src/config.c
```diff
@@ -68,22 +70,23 @@ config_tx(fido_dev_t *dev, uint8_t subcmd, cbor_item_t **paramv, size_t paramc,
...
 			fido_log_debug("%s: config_prepare_hmac", __func__);
 			goto fail;
 		}
-
-		/* If available, prefer cached PUAT */
-		if (fido_dev_puat_blob(dev) == NULL &&
-		    (r = fido_do_ecdh(dev, &pk, &ecdh, ms)) != FIDO_OK) {
-			fido_log_debug("%s: fido_do_ecdh", __func__);
-			goto fail;
+		if (!token) {
```
martelletto: nit: == NULL

src/cred.c
```diff
@@ -197,29 +197,17 @@ decrypt_hmac_secret(const fido_dev_t *dev, fido_cred_t *cred,
...
-
 int
 fido_dev_make_cred(fido_dev_t *dev, fido_cred_t *cred, const char *pin)
 {
-	fido_blob_t	*ecdh = NULL;
-	es256_pk_t	*pk = NULL;
-	int		 ms = dev->timeout_ms;
-	int 		 r;
+	const fido_blob_t	*token = NULL;
+	fido_blob_t	         token_store;
```
martelletto: same here

src/credman.c
```diff
@@ -124,17 +124,18 @@ static int
...
     const char *rp_id, fido_opt_t uv, int *ms)
 {
-	fido_blob_t	 f;
-	fido_blob_t	*ecdh = NULL;
-	fido_blob_t	 hmac;
-	es256_pk_t	*pk = NULL;
-	cbor_item_t	*argv[4];
-	const uint8_t	 cmd = credman_get_cmd(dev);
-	int		 r = FIDO_ERR_INTERNAL;
+	const fido_blob_t	*token = fido_dev_puat_blob(dev);
```
martelletto: same here

src/pin.c
```diff
@@ -340,46 +340,63 @@ cmd_to_perm(uint8_t cmd)
...
+		}
+		ecdh = ecdh_local;
+		pk = pk_local;
+	}
 
-	return (uv_token_wait(dev, perm, pin, ecdh, pk, rpid, token, ms));
+	r = uv_token_wait(dev, perm, pin, ecdh, pk, rpid, token, ms);
+fail:
+	es256_pk_free(&pk_local);
+	fido_blob_free(&ecdh_local);
```
martelletto: won't this free the memory backing `ecdh` and `pk`?

martelletto (commented):
Left a couple of comments; LGTM otherwise. If `tools/test.sh` is passing, let's make sure it still passes with the changes.

src/pin.c
LDVG: The `_local` variants are only used if both arguments are NULL (and consequently not required when this function exits).

martelletto: oh, I see. I thought `ecdh` and `pk` were both returned to the caller.

LDVG:
> Left a couple of comments

Thank you, tweaked accordingly.

> If tools/test.sh is passing, let's make sure it still passes with the changes.

It is, and it does!

src/pin.c
LDVG: I renamed these with a `tmp_` prefix instead to clarify.

dacav (approved):
🪓 Axe. You have mine.

PR #992, opened 2026-07-17
cmake: specify CMAKE_* variables as CACHE

dacav:
While keeping current default, honours the user choices.


CMakeLists.txt
```diff
@@ -26,9 +26,9 @@ if(CHECK_PIE_SUPPORTED)
 endif()
 
 set(CMAKE_POSITION_INDEPENDENT_CODE ON)
-set(CMAKE_COLOR_MAKEFILE OFF)
-set(CMAKE_VERBOSE_MAKEFILE ON)
-set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
+set(CMAKE_COLOR_MAKEFILE          OFF CACHE BOOL "Color output during build")
```
akallabeth, line 28: * why leave out `CMAKE_POSITION_INDEPENDENT_CODE`?
* keeping style with below, `option` would be the better choice.

dacav: > * why leave out `CMAKE_POSITION_INDEPENDENT_CODE`?

Actually, I think it can be removed: it is not affecting libfido2.so (for which PIC is already implied) and I don't think it serves any need.

> * keeping style with below, `option` would be the better choice.

I don't think so. Quoting <https://cmake.org/cmake/help/latest/command/option.html#command:option>:
> If <variable> is already set as a normal or cache variable, then the command does nothing (see policy [CMP0077](https://cmake.org/cmake/help/latest/policy/CMP0077.html#policy:CMP0077)).

So, I tried to use `option` instead, but then redefining a default value doesn't work.
For example, `option(CMAKE_COLOR_MAKEFILE "..." OFF)` is not effective, since the default is already `ON`, as defined by CMake.

PR #998, opened 2026-09-18
Allow wrapping sequence numbers for continuation frames 

balfanz:
The CTAP HID binding spec states that each continuation frame carry a sequence number, which uses 7 bits. This caps the number of possible continuation frames at 128, or (given frame sizes and space reserved for payload bits) the maximum size of a CTAP HID message at 7609 bytes.

There are larger payloads, however (e.g., PQC algs). The BLE HID spec, and other CTAP HID implementations allow the sequence number to wrap around. This still allows us to detect dropped or spurious frames, but removes the limit of 128 continuation frames. (We're still capping the processing of frames at payload_len.)

This change brings libfido2 in line with those implementations.


dacav:
Hi @balfanz,

Thanks for your contribution. We have some in progress modifications in this area, so we will integrate your patch in it, and keep you posted.

src/io.c
```diff
@@ -262,7 +262,14 @@ rx(fido_dev_t *d, uint8_t cmd, unsigned char *buf, size_t count, int *ms)
...
 
-	for (int seq = 0; r < payload_len; seq++) {
+	/*
+	 * The continuation frame sequence number is 7 bits wide. Strict
+	 * spec adherence would dictate capping it at 127. But larger
+	 * payloads are possible (e.g., PQC algorithms). We're therefore
+	 * aligning with BLE framing, Chromium implementation, and emerging
+	 * consensus in FIDO to allow wrapping of the sequence number.
+	 */
+	for (uint8_t seq = 0; r < payload_len; seq = (seq + 1) % 128) {
```
dacav, line 272: This works for me, but it gives a compiler warning due to `(seq + 1)` being promoted to `int`. May I suggest a casting, or simply to use `unsigned int`?

```suggestion
	for (unsigned int seq = 0; r < payload_len; seq = (seq + 1) % 128) {
```

martelletto:
`unsigned int` makes sense, but note that `tx()` already uses `uint8_t`. If we decide to stick with `uint8_t`, we could change `seq + 1` to `seq + 1U` instead of casting.

dacav:
@balfanz 
I created a branch named `pqc-algo` on the Yubico/libfido2 github repo. It is meant to collect all the changes related to pq algorithms. Could you please use that as merge target, once the warning is sorted out?

Thanks :)

